PR #5, opened 2019-03-12
Made :bb_runner_layer and :passwd_layer public

mmikitka:
See https://github.com/buildbarn/bb-remote-execution/issues/4


EdSchouten:
Thanks!

PR #8, opened 2019-04-27
Port to windows

### LINKED ISSUE #7: Windows worker
Hi, not sure if this is an issue but I didn't found anything on the web.
I basically understand how buildbarn is working. my question is:
how is the procedure to setup a build cluster having windows/Linux workers?
Is build barn meant to be running a windows worker?
(currently trying to compile the components on windows but it fails with a missed BUILD file in 
org_golang_google_grpc module)

moroten:
Apart from the three changes in this pull request, `local_directory.go` in `bb-storage` needs to be ported to Windows. See also buildbarn/bb-storage/pull/8.

This partly fixes #7.

### DIFF
--- a/cmd/bb_runner/main.go
+++ b/cmd/bb_runner/main.go
@@ -49,13 +49,16 @@ func main() {
 	s := grpc.NewServer()
 	runner.RegisterRunnerServer(s, runnerServer)
 
-	if err := os.Remove(runnerConfiguration.ListenPath); err != nil && !os.IsNotExist(err) {
-		log.Fatalf("Could not remove stale socket %#v: %s", runnerConfiguration.ListenPath, err)
+	if runnerConfiguration.ListenNetwork == "unix" {
+		if err := os.Remove(runnerConfiguration.ListenPath); err != nil && !os.IsNotExist(err) {
+			log.Fatalf("Could not remove stale socket %#v: %s", runnerConfiguration.ListenPath, err)
+		}
 	}
 
-	sock, err := net.Listen("unix", runnerConfiguration.ListenPath)
+	sock, err := net.Listen(runnerConfiguration.ListenNetwork, runnerConfiguration.ListenPath)
 	if err != nil {
-		log.Fatalf("Failed to create listening socket %#v: %s", runnerConfiguration.ListenPath, err)
+		log.Fatalf("Failed to create %#v listening socket %#v: %s",
+			runnerConfiguration.ListenNetwork, runnerConfiguration.ListenPath, err)
 	}
 	if err := s.Serve(sock); err != nil {
 		log.Fatal("Failed to serve RPC server: ", err)
--- a/pkg/environment/local_execution_environment.go
+++ b/pkg/environment/local_execution_environment.go
@@ -66,9 +66,9 @@ func (e *localExecutionEnvironment) Run(ctx context.Context, request *runner.Run
 	if len(request.Arguments) < 1 {
 		return nil, status.Error(codes.InvalidArgument, "Insufficient number of command arguments")
 	}
-	cmd := exec.CommandContext(ctx, request.Arguments[0], request.Arguments[1:]...)
-	// TODO(edsch): Convert workingDirectory to use platform
-	// specific path delimiter.
+	// Forward slash in the executable on Windows is interpreted as command argument. Convert to backslash.
+	executable := strings.ReplaceAll(request.Arguments[0], "/", string(filepath.Separator))
+	cmd := exec.CommandContext(ctx, executable, request.Arguments[1:]...)
 	cmd.Dir = filepath.Join(e.buildPath, request.WorkingDirectory)
 	for name, value := range request.EnvironmentVariables {
 		cmd.Env = append(cmd.Env, name+"="+value)
--- a/cmd/bb_worker/BUILD.bazel
+++ b/cmd/bb_worker/BUILD.bazel
@@ -4,7 +4,10 @@ load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
 go_library(
     name = "go_default_library",
-    srcs = ["main.go"],
+    srcs = ["main.go"] + select({
+        "@io_bazel_rules_go//go/platform:windows_amd64": ["clear_umask_windows.go"],
+        "//conditions:default": ["clear_umask_unix.go"],
+    }),
     importpath = "github.com/buildbarn/bb-remote-execution/cmd/bb_worker",
     visibility = ["//visibility:private"],
     deps = [
--- a/cmd/bb_worker/clear_umask_unix.go
+++ b/cmd/bb_worker/clear_umask_unix.go
@@ -1,0 +1,12 @@
+// +build !windows
+
+package main
+
+import (
+	"syscall"
+)
+
+// Calls syscall.Umask(0) on non-Windows systems.
+func clearUmask() {
+	syscall.Umask(0)
+}
--- a/cmd/bb_worker/clear_umask_windows.go
+++ b/cmd/bb_worker/clear_umask_windows.go
@@ -1,0 +1,7 @@
+// +build windows
+
+package main
+
+// Calls syscall.Umask(0) on non-Windows systems.
+func clearUmask() {
+}
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -8,7 +8,6 @@ import (
 	_ "net/http/pprof"
 	"net/url"
 	"os"
-	"syscall"
 	"time"
 
 	re_blobstore "github.com/buildbarn/bb-remote-execution/pkg/blobstore"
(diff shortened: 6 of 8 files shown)


cmd/bb_runner/main.go
```diff
@@ -18,6 +18,7 @@ func main() {
 	var tempDirectoriesList util.StringList
 	var (
 		buildDirectoryPath = flag.String("build-directory", "/worker/build", "Directory where builds take place")
+		listenNetwork      = flag.String("listen-network", "unix", "Network type to listen on: tcp, tcp4, tcp6, unix or unixpacket.")
```
EdSchouten: The examples given here are not exhaustive, right? Even if the current version of Go only supports these, the actual list is not something we can influence. Maybe better to put 'e.g.' somewhere in the description to reflect that.

pkg/environment/local_execution_environment.go
```diff
@@ -66,9 +67,12 @@ func (e *localExecutionEnvironment) Run(ctx context.Context, request *runner.Run
 	if len(request.Arguments) < 1 {
 		return nil, status.Error(codes.InvalidArgument, "Insufficient number of command arguments")
 	}
-	cmd := exec.CommandContext(ctx, request.Arguments[0], request.Arguments[1:]...)
-	// TODO(edsch): Convert workingDirectory to use platform
-	// specific path delimiter.
+	executable := request.Arguments[0]
+	if runtime.GOOS == "windows" {
+		// Forward slash in the executable on Windows is interpreted as command argument. Convert to backslash.
```
EdSchouten: So this means that execute requests always use executable names containing forward slashes, even when targeting Windows? Ouch! Do we know whether this is something intentional on Bazel's side? Is there some existing option to switch the path separator convention here?

General nit: instead of conditionalizing this on Windows, what about doing this always? That keeps the code compact. The same likely holds for the working directory, right?

moroten: > So this means that execute requests always use executable names containing forward slashes, even when targeting Windows? Ouch! Do we know whether this is something intentional on Bazel's side? Is there some existing option to switch the path separator convention here?

I haven't looked into what Bazel is meant to do and cannot find any documentation during a quick search. [`actions.run`](https://docs.bazel.build/versions/master/skylark/lib/actions.html#run) has a separate `executable` argument, so that might be part of it. Most use cases copes with forward slash on Windows, but apparently not everywhere. Probably need to do a proper check with a C api call to see if the problem is Bazel or Go not interacting correctly with `CreateProcess()`.

> General nit: instead of conditionalizing this on Windows, what about doing this always? That keeps the code compact. The same likely holds for the working directory, right?

Will fix the conditionalizing. Working directory does not need any handling, that one works.

cmd/bb_runner/main.go
EdSchouten: The latest version of Buildbarn would require no more (or at least less) patching on this end. It is possible to now use UNIX sockets or TCP sockets for any purpose. As in, bb-worker -> bb-runnner can now use TCP as well.

YngveNPettersen:
A few observations after I have been testing this patch on Windows (after resolving the conflicts and compile issues):

- I am seeing problems with removing the work directories due to stderr (at least) being in use by another application. I have not yet investigated this, so I do not yet know which application is locking it.
- I would recommend against using hardlinking on Windows, I have had problems with the error "An attempt was made to create more links on a file than the file system supports." being reported, after work directories have not been cleaned up after use, probably due to file lock issue above (in the most recent case there was >1000 directories consuming 60+ GB disk space)
- The hardlinking has also encountered the error "Cannot create a file when that file already exists.", I "fixed" that by always removing the target before re-linking it. Not sure what is causing this.
- I therefore suggest using file copy instead of linking on Windows
- I disabled the UNC code, since I could not get the application working with it,

moroten:
The main message is that FUSE is the way to go for the Buildbarn worker. However, this filesystem still needs to work.

> * I am seeing problems with removing the work directories due to stderr (at least) being in use by another application. I have not yet investigated this, so I do not yet know which application is locking it. Update: It is bb_worker itself that locked stderr, stdout, and at least the output object file, preventing it from deleting the build directory. Further update, the problem seems to be that localBuildExecutor.Execute() created the build directory environment, and registrers a defer to release it on exit. Before it exits, however, Execute registers uploads of several files, but these are not flushed before the caller storageFlushingBuildExecutor.Execute calls the flush function _after_ localBuildExecutor has completed, which leads to a race condition: The directory cannot be deleted until the upload queue has been flushed (which will cause the files to be closed), but that cannot be done until after the directory is supposed to be deleted.

Comment from @EdSchouten in https://github.com/buildbarn/bb-storage/pull/8:

> > _If we're going into the direction of using these functions, we could at the same time tackle the issue you experienced. By opening files using FILE_SHARE_DELETE, we can hold on to references after deletion. Some GSoC project, Boost AFIO, has some interesting notes on that:
> > https://boostgsoc13.github.io/boost.afio/doc/html/afio/quickstart/mini_programs/filesystem_races.html
> > https://boostgsoc13.github.io/boost.afio/doc/html/afio/FAQ/deleting_open_files.html_

Basically, the Buildbarn has to move the output files up one directory, with random file name, before removing the folder. This should be doable.

> * I would recommend against using hardlinking on Windows, I have had problems with the error "An attempt was made to create more links on a file than the file system supports." being reported, after work directories have not been cleaned up after use, probably due to file lock issue above (in the most recent case there was >1000 directories consuming 60+ GB disk space)

Interesting. The problem I had was instead that hardlinking 10000 files per action took 30s, far to much. Therefore, I've started some work using on https://github.com/billziss-gh/cgofuse/ and got a stand alone file system running which supports an in memory tree with metadata where files are written to a single folder using inode as file name. Integrating it into Buildbarn would mean loading files on demand and opening them directly from the cache directory when reading.

> * The hardlinking has also encountered the error "Cannot create a file when that file already exists.", I "fixed" that by always removing the target before re-linking it. Not sure what is causing this.

Not sure either. Have only run a limited proof of concept test.

> * I therefore suggest using file copy instead of linking on Windows

I assume file copying is slower than hard linking. Anyway, FUSE is the way to go in my perspective.

> * I disabled the UNC code, since I could not get the application working with it,

Ok. In https://github.com/buildbarn/bb-storage/pull/8, @EdSchouten mentions using Nt-API instead of Win32 API for directory handles instead of paths. Maybe that will solve the problem. Also, when using FUSE, the problem of long paths should disappear.

>> _Just like *at() on POSIX systems, Windows NtCreateFile() can also open files using directory file descriptors. It turns out there is a Go package that provides bindings for these system calls:
>> https://github.com/hillu/go-ntdll_

EdSchouten:
> Basically, the Buildbarn has to move the output files up one directory, with random file name, before removing the folder. This should be doable.

Exactly. Better said: it needs to be moved to the root of the hierarchy. You could make it so that `localDirectory` on Windows keeps track of two handles: one for the directory itself and one for the directory where deleted files need to be placed.

`NewLocalDirectory()` would create a `localDirectory` where both those handles are the same. When calling `Directory.Enter()` on such a directory, you will get a `localDirectory` with two different internal handles.

YngveNPettersen:
> The problem I had was instead that hardlinking 10000 files per action took 30s, far to much.

Hmmm, this may be old, and possibly obsolete, information, but I have previously had performance issues on Windows when a single folder (like the buildbarn cache ) have a lot of files. My recommendation is to break such big allocations up into multiple folders, e.g grouped by the first two characters in the name.

YngveNPettersen:
I may have found the cause of the "number of hardlinks" failure : If looks like if there are more than ~50K files in the cache directory, that error will occur. By storing the cache files in subdirs based on the two first characters in the hash key the problem went away.

I also suspect that the "can't link because file already exists" problem is due to multiple threads fetching the same file at almost the same time, leading to a race condition. I think downloads should be cross-thread coordinated, so that only one is downloading a file, the others wait for it to be downloaded (and do other download work in the mean time)

YngveNPettersen:
I am continuing my testing on Windows and one major problem at present is that the build work dirs can't be deleted because they are open in the compiler, even if the task is long finished.

My current idea about the cause of this is that if the compiler is opening any one of the locations of the hard-linked file, then it locks *all* of them, not just the location it is opening. This then prevent the deletion of the work directories until no compilers are running

moroten:
I was mostly blaming our anti-virus software for locked files. Anyway, should be solved by using FUSE.

YngveNPettersen:
The normal error messages does not say which application is locking the file, but if you try to delete the files using Windows Explorer, at least on Win 10, the error message will tell you which application is locking the file, although not which process ID.

I've now added a throttle system that will not create new work directories if the number of active directories is more than 4 times the concurrency limit, in such cases I made it sleep for a second, then checking the number again, until it is below twice the concurrency limit. I separate out the dir clean operation as a small go routine for each build directory that can't be deleted.

I may have noticed another issue with the file locking: if the action is restarted for some reason, before the directory is cleared, then there is a risk that the file population operation fails with "file exists", I am trying to resolve this by increasing the length of the build dir name from the hash value if the directory already exists; that seems to have worked, since I have now finally been able to do a full build of our project.

EdSchouten:
Closing this due to inactivity. Some of the features that were requested in this patch (e.g., UNIX socket vs. TCP socket) have already been generalised. Feel feel to reply to this PR or file a new one for individual features if desired!

moroten:
I'm fine with closing the Windows porting. I've been inactive by prioritizing getting our Linux builds working first. Windows users have to stick to remote cache or use Wine during remote execution, when it works.

PR #9, opened 2019-05-10
Add bb_runner_bare_container

ailiev:


ailiev:
I had a quick chat with my rubber duck and realized the existing images can transmit the binary just as well, so no need for this.

PR #10, opened 2019-05-10
Allow specifying the remote execution digest fn on the scheduler cmd line

ailiev:
In case the bazel has to run with a function other than SHA256 - it checks that the remote executor announces the same function.

### DIFF
--- a/cmd/bb_scheduler/main.go
+++ b/cmd/bb_scheduler/main.go
@@ -34,7 +34,20 @@ func main() {
 		log.Fatal(http.ListenAndServe(schedulerConfiguration.MetricsListenAddress, nil))
 	}()
 
-	executionServer, schedulerServer := builder.NewWorkerBuildQueue(util.DigestKeyWithInstance, schedulerConfiguration.JobsPendingMax)
+	var execDigestFunc remoteexecution.DigestFunction
+	switch schedulerConfiguration.ExecutionDigestFunction {
+	case "sha256":
+		execDigestFunc = remoteexecution.DigestFunction_SHA256
+	case "sha1":
+		execDigestFunc = remoteexecution.DigestFunction_SHA1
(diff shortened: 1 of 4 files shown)


mickael-carl:
Could you rebase those changes on top of master?

ailiev:
> Could you rebase those changes on top of master?

Done

cmd/bb_scheduler/main.go
```diff
@@ -34,7 +34,20 @@ func main() {
 		log.Fatal(http.ListenAndServe(schedulerConfiguration.MetricsListenAddress, nil))
 	}()
 
-	executionServer, schedulerServer := builder.NewWorkerBuildQueue(util.DigestKeyWithInstance, schedulerConfiguration.JobsPendingMax)
+	var execDigestFunc remoteexecution.DigestFunction
+	switch schedulerConfiguration.ExecutionDigestFunction {
+	case "sha256":
```
mickael-carl: For this couldn't you use rather the [facilities provided by protobuf](https://developers.google.com/protocol-buffers/docs/reference/go-generated#enum)? For reference, you could look at [the generated code](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.pb.go#L68). This would avoid using strings here.

EdSchouten: Or: Instead of using a string, you could use the enum type directly inside the .proto file. That way they are guaranteed to remain in sync.

ailiev: yup, I'm trying to pull in the remote-apis enum . It's resisting but I hope to subdue it soon.

EdSchouten:
Closing this due to inactivity. Feel free to reply to this PR or file a new one in case you're still interested in adding this feature!

Issue #2, opened 2019-02-18
Use the Remote Worker API instead of a custom protocol... or not?


EdSchouten:
At the Bloomberg Build Event and FOSDEM earlier this month, we chatted about convergence of remote worker protocols between Buildbarn and BuildGrid. Even though RWAPI would be the most obvious choice, it was mentioned that the authors of this protocol have less interest in using it externally than initially anticipated.

@buchgr stated that he is going to see whether he can get RWAPI moved to https://github.com/bazelbuild/remote-apis. When imported, we should hold discussions to see whether the protocol can be simplified. Once that's done, we should alter bb-scheduler and bb-worker to use that protocol instead.

Old tracking bug: https://github.com/EdSchouten/bazel-buildbarn/issues/7

jjardon:
Hi, is there any further movement on this area?

Seems the RWAPI is not at https://github.com/bazelbuild/remote-apis yet

EdSchouten:
Hi Javier,

@buchgr mentioned earlier this week he's still dealing with Google internal politics getting this moved over. He pinged some of the other people involved once again, so let's hope some progress is achieved one of these days/weeks.

finnball:
if it does happen, do you plan on implementing soon or is it something lower priority for you?

buchgr:
> let's hope some progress is achieved one of these days/weeks.

My guess is that it will be months but the most current version (the one that needs to go through review) of the API is https://github.com/googleapis/googleapis/tree/master/google/devtools/remoteworkers/v1test2 and so we could still make progress by implementing this version, as I'd expect there to only be very minor changes being requested during the review.

EdSchouten:
Discussion going on here: https://groups.google.com/forum/#!msg/remote-execution-apis/QLJvYKd5_gw/8rVVbLUmBgAJ

EdSchouten:
The outcome of all of this is that Buildbarn now has an improved protocol that works incredibly well. It is up to other implementations to decide whether they want to adopt something it too.

PR #13, opened 2019-05-31
WIP: Implement polling based RWAPI

finnball:
This aims to implement a polling based API for Buildbarn as detailed in #2 .

The API is detailed [here](https://github.com/finnball/bb-remote-execution/blob/finn/reapiv2/pkg/proto/worker/remote_worker.proto)

Still to do:
- Unittests
- Hash the worker_id
- Think of a better method of assigning a job id [here](https://github.com/buildbarn/bb-remote-execution/pull/11/commits/c152f8fd934eaefb29831d2c6041bd2e4cae66be#diff-14ece2b722d93fb123ddc7a12dc7781eR269)


finn-ball:
Scrap this branch as I no longer have access to it.

EdSchouten:
#28 introduced the new scheduler, so there is no need for this PR anymore. Thanks for sending it in regardless!

PR #16, opened 2019-07-22
Make the worker cache size configurable

moroten:
This change removes some hard coded constants from the code. Now it will be possible set the worker cache larger than 1 GiB.

### DIFF
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -84,8 +84,9 @@ func main() {
 			cas.NewBlobAccessContentAddressableStorage(
 				re_blobstore.NewExistencePreconditionBlobAccess(contentAddressableStorageBlobAccess),
 				workerConfiguration.MaximumMessageSizeBytes),
-			util.DigestKeyWithoutInstance, cacheDirectory, 10000, 1<<30),
-		util.DigestKeyWithoutInstance, 1000)
+			util.DigestKeyWithoutInstance, cacheDirectory,
+			int(workerConfiguration.MaximumCacheFileCount), int64(workerConfiguration.MaximumCacheSizeBytes)),
+		util.DigestKeyWithoutInstance, int(workerConfiguration.MaximumMemoryCachedDirectories))
 	actionCache := ac.NewBlobAccessActionCache(actionCacheBlobAccess)
 
 	// Create connection with scheduler.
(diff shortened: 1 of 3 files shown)


pkg/configuration/bb_worker/configuration.go
```diff
@@ -28,11 +28,20 @@ func setDefaultWorkerValues(workerConfiguration *pb.WorkerConfiguration) {
 	if workerConfiguration.CacheDirectoryPath == "" {
 		workerConfiguration.CacheDirectoryPath = "/worker/cache"
 	}
+	if workerConfiguration.MaximumCacheFileCount == 0 {
+		workerConfiguration.MaximumCacheFileCount = 10000
+	}
+	if workerConfiguration.MaximumCacheSizeBytes == 0 {
+		workerConfiguration.MaximumCacheSizeBytes = 1024 ^ 3
```
mickael-carl: Careful here, `1024 ^ 3` will not do what you expect it to do!

cmd/bb_worker/main.go
```diff
@@ -13,8 +13,8 @@ import (
 
 	re_blobstore "github.com/buildbarn/bb-remote-execution/pkg/blobstore"
 	"github.com/buildbarn/bb-remote-execution/pkg/builder"
-	"github.com/buildbarn/bb-remote-execution/pkg/configuration/bb_worker"
 	cas_re "github.com/buildbarn/bb-remote-execution/pkg/cas"
+	configuration "github.com/buildbarn/bb-remote-execution/pkg/configuration/bb_worker"
```
mickael-carl: What is the rationale of renaming this import?

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -19,6 +19,15 @@ message WorkerConfiguration {
   // Directory where build input files are cached. Defaults to "/worker/cache".
   string cache_directory_path = 4;
 
+  // Maximum number of files in the cache. Defaults to 10000.
+  uint64 maximum_cache_file_count = 10;
+
+  // Maximum total size of the cache in bytes. Defaults to 1024^3, i.e. 1 GiB.
```
mickael-carl: Maybe change the comment here as well to reflect the 1024^3 change.

pkg/configuration/bb_worker/configuration.go
moroten: Such a mistake! Don't know why. Thank's for spotting.

cmd/bb_worker/main.go
moroten: My VS Code auto-formatter did that. Have reverted it.

mickael-carl:
Thanks!

PR #17, opened 2019-08-03
Split ContentAddressableStorage interface into reader and writer

moroten:
It is possible to split the `ContentAddressableStorage` interface into a reader and a writer. `ReadWriteDecouplingContentAddressableStorage` already exists and is using one reader and one writer, so this split is a natural refactoring.

### DIFF
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -137,23 +137,24 @@ func main() {
 			// Per-worker separate writer of the Content
 			// Addressable Storage that batches writes after
 			// completing the build action.
-			contentAddressableStorageWriter, contentAddressableStorageFlusher := re_blobstore.NewBatchedStoreBlobAccess(
+			blobAccessWriter, contentAddressableStorageFlusher := re_blobstore.NewBatchedStoreBlobAccess(
 				re_blobstore.NewExistencePreconditionBlobAccess(contentAddressableStorageBlobAccess),
 				util.DigestKeyWithoutInstance, 100)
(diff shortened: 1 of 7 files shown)


EdSchouten:
The reason why I never merged this change was not because I was necessarily against it, but because I first wanted to see how the code around it evolved. Now that some time has passed, I think there's another option we should pursue: simply remove the ContentAddressableStorage interface.

In the past we had an ActionCache interface that similarly wrapped around BlobAccess to provide a typed API. This interface was removed when we added pkg/blobstore/buffer, because Buffer has a ToActionResult() function that can easily be used to convert to ActionResult messages.

I am currently working on a change to replace `Buffer.ToActionResult()` with a `Buffer.ToProto()`. Similarly, Buffer objects can also be constructed from arbitrary Protobuf messages. Once that lands, you can argue that all functions except `GetFile()` and `PutFile()` become redundant. With almost the same effort, you could just call `BlobAccess.Get()`, followed by `Buffer.ToProto()`.

`GetFile()` is also a bit wacky, because you wouldn't want to use this when virtualising the input root. In that case you'd likely want to use `BlobAccess.Get()` and `Buffer.ReadAt()`. You do want to have it for the non-virtualised case, but that could just as well be placed in a dedicated `FileDownloader` interface or something. `HardlinkingContentAddressableStorage` would then become a `HardlinkingFileDownloader`.

Long story short, let's not pursue a change like this. We'd better leave things as is and revisit this by the time we have `Buffer.ToProto()` (and something like `FileDownloader`). Thanks for your patch regardless!

PR #25, opened 2019-10-19
Add scheduler build queue metrics

Qinusty:
This adds prometheus metrics for tracking the current size of the queue in a scheduler. Metrics are exposed to show the current size of the queue and the individual push/pop requests the scheduler makes.


EdSchouten:
Let's close this PR, as the new scheduler introduced in #28 exposes a bunch of interesting stats.

PR #27, opened 2019-12-04
WIP: Polling API

finn-ball:
I believe this can be simplified further. I'll leave it in WIP until simplified and updated with unittests.


EdSchouten:
With #28 landed, we now use a new worker protocol + scheduler. Let's mark this as fixed/closed!

Issue #24, opened 2019-09-24
Unable to access GCS due to missing certs


henryaj:
Hi!

We're seeing problems when using GCS as a storage backend; specifically errors like:

```
Get https://storage.googleapis.com/<bucketname>/buildbarn-cas/<some-hash>:
oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: x509:
certificate signed by unknown authority
```

Is this fixable with a simple `apt-get install ca-certificates` or similar? Notably not all the images seem to have this problem; specifically we see this error when doing remote execution (as opposed to remote caching, which works fine).

EdSchouten:
The container images now use the rules_docker go_image() as their base image. This means they now include CAs properly. Marking this as closed. Thanks for reporting this issue!

Issue #3, opened 2019-03-08
More robust and scalable scheduler


finnball:
I'm looking into large scale deployments of remote execution with buildbarn. I've identified the scheduler as an area we'd like to have more robust and scalable . The problems I have with the scheduler are:

- Scheduler goes down, we lose all statuses of jobs
- Not easy to elastically scale the max number of jobs
- Not easy to make clones of schedulers with pods of workers and make `frontends` (or storage daemons) aware of new schedulers

I propose that Buildbarn make it possible to remove the storage of the job and its metadata from the actual implementation of the scheduling algorithm. This would allow another service to take care of the fallback incase the service goes down and it needs to repopulate.

Each `Backend` below could be a different scheduling algorithm. It could, for example, query for jobs with specific metadata, perhaps even depending on the workers it has available. 

Have you any thoughts on this problem, comments, queries or input? I'd be especially interested if you had any ideas or suggestions on what to use with the `JobStorage`.

```
                                                                      Potential Design

                                                      +-------------+  +-------------+  +-------------+
                                                      |             |  |             |  |             |
               Current Design                         |  Frontend   |  |  Frontend   |  |  Frontend   |
                                                      |             |  |             |  |             |
     +-------------+    +-------------+               +------+------+  +------+------+  +------+------+
     |             |    |             |                      |                |                |
     |  Frontend   |    |  Frontend   |                      |                |                |
     |             |    |             |             +--------+----------------+----------------+--------+
     +-----+-------+    +------+------+             |                                                   |
           |                   |                    |                                                   |
           +-------------------+                    |                   JobStorage                      |
           |                   |                    |                                                   |
    +------+--------+  +-------+-------+            |                                                   |
    |               |  |               |            +--------+----------------+----------------+--------+
    |   Scheduler   |  |   Scheduler   |                     |                |                |
    |               |  |               |                     |                |                |
    +-+---------+---+  +---+---------+-+              +------+------+  +------+------+  +------+------+
      |         |          |         |                |             |  |             |  |             |
      |         |          |         |                |  Backend    |  |  Backend    |  |  Backend    |
+-----+--+ +----+---+  +---+----+ +--+-----+          |             |  |             |  |             |
|        | |        |  |        | |        |          +-+---------+-+  +------+------+  +--+--------+-+
| Worker | | Worker |  | Worker | | Worker |            |         |           |            |        |
|        | |        |  |        | |        |            |         |           |            |        |
[...]

Issue #1, opened 2019-02-18
Scheduler endpoint should show list of workers and currently running actions


EdSchouten:
Right now the scheduler doesn't expose any information about its running state. We should add some kind of status page that shows at least the following information:

- The list of workers connected to the scheduler,
- Which actions these workers are currently executed,
- The list of actions that is pending to be executed,
- Optional: a list of recently completed actions.

We can simply link to bb-browser to show meaningful information about the actions.

Old tracking bug: https://github.com/EdSchouten/bazel-buildbarn/issues/12

EdSchouten:
All of this is present nowadays.

PR #28, opened 2019-12-24
Initial Open Source release

EdSchouten:
This PR should fix the following issues: #1 #2 #3 #13 #24 #25 #27

This PR brings in a new custom Remote Worker protocol. This protocol is aware of platform properties and allows safe resumption of build actions in case of network hiccups. The new scheduler that implements this protocol has a fancy web UI that supports basic interaction with the system as well (killing operations, draining workers, etc.).


pkg/builder/in_memory_build_queue.go
```diff
@@ -0,0 +1,1489 @@
...
+		return err
+	}
+
+	bq.enter(bq.clock.Now())
+	defer bq.leave()
+
+	pq, ok := bq.platformQueues[platformKey]
+	if !ok {
+		return status.Errorf(codes.FailedPrecondition, "No workers exist for instance %#v platform %s", platformKey.instance, platformKey.platform)
+	}
```
moroten: The implementation of the platform logic is well done, but matches exact platform properties only. One suggested RW API, [API Design: Remote Workers API](https://docs.google.com/document/d/1s_AzRRD2mdyktKUj2HWBn99rMg_3tcPvdjx3MPbFidU/edit#heading=h.1lt0k9rr0yu4) under the chapter _Configurations_, suggests possibility to run on any worker that fulfills **at least** the properties advertised by the worker. One example is that one action requires SSE4 but another one does not, but both can run on the same worker. I have also seen a good use of advertising for example `DockerImage=` from the worker side, to support a worker that can accept actions for any image.

Changing this logic is a bit cumbersome and can be done in a separate pull request. I'm thinking in letting `Execute()` populate the `platformQueues` map and let workers poll multiple `platformQueues` in `Synchronize()`.

pkg/environment/action_digest_subdirectory_manager.go
```diff
@@ -41,9 +39,14 @@ func (em *actionDigestSubdirectoryManager) Acquire(actionDigest *util.Digest) (M
...
-	// Create build directory within.
+	// Create build directory within. Only use a small number of
+	// characters from the digest to ensure the absolute path of the
+	// build directory remains short. This avoids reaching PATH_MAX
+	// and sockaddr_un::sun_path size limits for stronger digest
+	// functions. 16 characters is more than sufficient to prevent
+	// collisions.
 	buildDirectory := environment.GetBuildDirectory()
-	subdirectoryName := actionDigest.GetKey(em.subdirectoryFormat)
+	subdirectoryName := actionDigest.GetHashString()[:16]
```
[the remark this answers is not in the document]
EdSchouten: Yeah, 16 characters is more than sufficient. Even in the case of hash collisions there's no real problem. It will trigger a retriable error, meaning Bazel will simply retry it. On clusters with a higher number of workers, there is a slim chance of subsequent failures.

pkg/builder/in_memory_build_queue.go
EdSchouten: You got that right! The first step was to implement just the basic ways of matching platforms. Subsequent improvements to add fuzzy ways of matching platform queues should be added later on.

The question becomes what an efficient lookup structure looks like, especially when the number of platform queues becomes large. Right now we simply stringify the platform properties and do a map lookup, which should scale well enough.

Doing subset/superset matching by default seems a bit risky in my case, as it means the matching criteria may change even if seemingly unrelated platform queues get created/removed from the system. But that's not a big problem if we get the UX right.

EdSchouten:
> I've skimmed through a good amount of the code and it looks generally good. No blocking issue from my side. Will probably build and run it during the first half of January.

Thanks! Considering that I've been using this code for quite some time without running into any major issues, I'll likely merge it before then. That said, this is already a big improvement over the code that's in master right now (that is known to be broken with the latest version of Bazel).

> I assume the FUSE file system for workers are coming in later pull requests.

Yes. My plan was to add it to this PR, but with the holiday season I didn't have time to extract all of it. These PRs do include several changes that are a necessity for FUSE support (e.g., bb-storage/pkg/blobstore/buffer and bb-remote-execution/pkg/filesystem).

PR #32, opened 2020-02-13
Add opencensus to bb_scheduler and bb_worker

arlyon:
This adds detailed span reporting to bb_scheduler and bb_worker. I opted for overreporting rather than underreporting.

The only questionable detail right now: I understand that bb prefers decorators to add logic however `opencensus` doesn't gracefully handle the worker 'pull' model whereby both bazel and the workers are making requests to the scheduler. This means that the span ID has to be manually appended onto the grpc response to the workers, and manually retrieved on the worker side. Beacuse this requires access to the headers, it's not possible to bundle this logic into a build executor.

Because this approach is still up in the air, the header impl is still a little crude (but effective) until I get feedback.

I also split up the load_build_executor into three stages to get better observability over the download, invoke, and upload phases.

image

Also, if there is any data in buildbarn you'd like annotated, I'm open to ideas.


pkg/opencensus/init.go
```diff
@@ -0,0 +1,46 @@
...
+	pb "github.com/buildbarn/bb-storage/pkg/proto/configuration/bb_storage"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.opencensus.io/plugin/ocgrpc"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/trace"
+	"go.opencensus.io/zpages"
+)
+
+// Initialize sets up Opentracing with Jaeger and a Prometheus exporter.
+func Initialize(configuration *pb.JaegerConfiguration, namespace string) {
```
EdSchouten: This function is likely not any different from the copy that's already in bb-storage, right? If so, be sure to throw this out and cal the one from bb-storage. bb-storage is intended to act as a utility library for the rest of bb-* anyway.

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -7,6 +7,7 @@ import "google/protobuf/duration.proto";
 import "pkg/proto/configuration/blobstore/blobstore.proto";
 import "pkg/proto/configuration/eviction/eviction.proto";
 import "pkg/proto/configuration/grpc/grpc.proto";
+import "pkg/proto/configuration/bb_storage/bb_storage.proto";
```
EdSchouten: Could you sort the imports alphabetically, please? Thanks!

pkg/blobstore/batched_store_blob_access.go
```diff
@@ -78,6 +79,12 @@ func (ba *batchedStoreBlobAccess) flushLocked(ctx context.Context) {
 		return
 	}
 
+	ctx, span := trace.StartSpan(ctx, "blobstore.BatchedStoreBlobAccess.upload")
```
EdSchouten: It looks like many of these tracing spans don't contain a lot of additional data compared to what is communicated over the prototype of this function. Could I persuade you to create a separate TracingBlobAccess, just like how we already have MetricsBlobAccess? Then we could patch up pkg/blobstore/configuration to automatically instantiate it all over the place.

pkg/blobstore/existence_precondition_blob_access.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
 	"github.com/buildbarn/bb-storage/pkg/util"
+	"go.opencensus.io/trace"
```
EdSchouten: Nit: current code style is to place an empty line when imports come from different domain names. Domains are also sorted alphabetically. So in this case, opencensus.io should go below golang.org.

pkg/builder/build_client.go
```diff
@@ -13,8 +13,13 @@ import (
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
 
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
+
+	"go.opencensus.io/trace"
```
EdSchouten: ... Just like how you're doing imports here. This is good!

pkg/cas/read_write_decoupling_content_addressable_storage.go
```diff
@@ -28,41 +29,71 @@ func NewReadWriteDecouplingContentAddressableStorage(reader cas.ContentAddressab
 }
 
 func (cas *readWriteDecouplingContentAddressableStorage) GetAction(ctx context.Context, digest *util.Digest) (*remoteexecution.Action, error) {
+	ctx, span := trace.StartSpan(ctx, "cas.RWDecoupling.GetAction")
```
EdSchouten: The same here: should we add a TracingContentAddressableStorage that acts as a decorator?

EdSchouten:
Oh yeah, just a random thing to ask before I forget:

So if I get it right, OpenCensus/OpenTracing/Jaeger/... generates a unique identifier for every unique flow of RPC traffic. This gets passed along as a token throughout the entire codebase. In the end individual services (bb-storage, bb-worker, etc.) report their part of the trace to a centralised service, so you can view it in a browser. Right?

Would it be somehow possible to let bb-worker store this token in the ActionResult before it gets stored in the Action Cache? What I'm getting at: bb-browser already stores some execution stats for build actions. Could we let bb-browser link directly to Jaeger, so that you can easily view a trace for a given build action?

EdSchouten:
I just merged these changes:

https://github.com/buildbarn/bb-storage/commit/fb0be44cf0e61fe98608748b47f16ee5049d142a
https://github.com/buildbarn/bb-remote-execution/commit/e44cc90aabb2f23091310de89e38f752bd8b6069

This should make it easier for you to add Jaeger support to all components.

EdSchouten:
Closing this due to inactivity. Be sure to reach out if you are continuing to work on this!

PR #35, opened 2020-02-20
Fix output directory creation to be relative to working directory

### LINKED ISSUE #18: Buildbarn creates output directories based on the root directory
Buildbarn sets output directories relative to root directory as opposed to the working directory.

This can be seen [here](https://github.com/buildbarn/bb-remote-execution/blob/master/pkg/builder/local_build_executor.go#L188).

The RE protos (v2) [specify](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L509) that the working directory as:
```
  // The paths are relative to the working directory of the action execution.
  // The paths are specified using a single forward slash (`/`) as a path
  // separator, even if the execution platform natively uses a different
  // separator. The path MUST NOT include a trailing slash, nor a leading slash,
  // being a relative path. The special value of empty string is allowed,
  // although not recommended, and can be used to capture the entire working
  // directory tree, including inputs.
...
repeated string output_directories = 4;

  // The working directory, relative to the input root, for the command to run
  // in. It must be a directory which exists in the input tree. If it is left
  // empty, then the action is run in the input root.
string working_directory = 6;
```

This becomes an issue when the working directory is not specified as the input root and the output is specified as a subdirectory of the input directory outside of the working directory.
```
build_directory
├── input_root
└── output_dir
```

### bb browser
Looking at an example action from a buildstream execution through the eyes of bb-browser shows the following [buildbarn-buildstream.pdf](https://github.com/buildbarn/bb-remote-execution/files/3526063/buildbarn-buildstream.pdf)

Buildstream specifies the working directory as `/buildstream/autotools/hello.bst/doc/amhello` whilst expecting buildbarn to provide `../../../../../` relative to the working directory as an output directory (which would be `/`).

Buildbarn seems to interpret this as `../../../../../` relative to `/`.

### Required change

Specified paths:
- Working directory is relative to input root
- Output directories are relative  to working directory

Current paths:
- Working directory is relative to input root
- Output directories are relative to input root

cphang99:
REv2 mandates that output directories are created relative to
the working directory. Currently buildbarn creates these directories
relative to the build directory.

Construct output directory paths that are relative to the build
directory, as required by the filesystem package, but that are
relative to the working directory, as required by REv2

Closes #18

### DIFF
--- a/pkg/builder/local_build_executor.go
+++ b/pkg/builder/local_build_executor.go
@@ -25,37 +25,6 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-type outputDirectory struct {
-	path string
-}
-
-func newOutputDirectory(workingDirectory string, pathDir string) outputDirectory {
-	// REv2 mandates that output directories are created relative to the
-	// working directory. However, directory semantics in buildbarn only
-	// permit directory traversal in the forward direction to avoid
-	// any possible escaping of the input root.
-	outputPath := workingDirectory
-	components := strings.FieldsFunc(pathDir, func(r rune) bool { return r == '/' })
-
-	for _, component := range components {
-		if component == "." {
-			continue
-		} else if component == ".." {
-			outputPath = path.Dir(outputPath)
-		} else {
-			outputPath = path.Join(outputPath, component)
-		}
-	}
-
-	return outputDirectory{
-		path: path.Clean(outputPath),
-	}
-}
-
-func (o outputDirectory) Path() string {
-	return o.path
-}
-
 type localBuildExecutor struct {
 	contentAddressableStorage cas.ContentAddressableStorage
 	environmentManager        environment.Manager
@@ -271,11 +240,10 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 	// Build rules generally expect the parent directories to already be
 	// there. We later use the directory handles to extract output files.
 	outputParentDirectories := map[string]filesystem.Directory{}
-	workingDirectory := command.WorkingDirectory
 	for _, outputDirectory := range command.OutputDirectories {
-		dirPath := newOutputDirectory(workingDirectory, path.Dir(outputDirectory)).Path()
+		dirPath := path.Dir(outputDirectory)
 		if _, ok := outputParentDirectories[dirPath]; !ok {
-			dir, err := be.createOutputParentDirectory(buildDirectory, dirPath)
+			dir, err := be.createOutputParentDirectory(buildDirectory, path.Join(command.WorkingDirectory, dirPath))
 			if err != nil {
 				attachErrorToExecuteResponse(response, err)
 				return response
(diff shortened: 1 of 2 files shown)


tmewett:
Thank you, this patch fixes the issues I had using CMake described in #18 :)

pkg/builder/local_build_executor.go
```diff
@@ -25,6 +25,37 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+type outputDirectory struct {
+	path string
+}
+
+func newOutputDirectory(workingDirectory string, pathDir string) outputDirectory {
```
EdSchouten: Isn't all of this the same as simply calling `path.Join(workingDirectory, pathDir)`

pkg/builder/local_build_executor.go
```diff
@@ -240,8 +271,12 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
...
 	outputParentDirectories := map[string]filesystem.Directory{}
+	workingDirectory := command.WorkingDirectory
+	if workingDirectory == "" {
+		workingDirectory = "."
+	}
 	for _, outputDirectory := range command.OutputDirectories {
-		dirPath := path.Dir(outputDirectory)
+		dirPath := newOutputDirectory(workingDirectory, path.Dir(outputDirectory)).Path()
 		if _, ok := outputParentDirectories[dirPath]; !ok {
 			dir, err := be.createOutputParentDirectory(buildDirectory, dirPath)
```
EdSchouten: What if you were to leave all of the existing bookkeeping intact (so still keep `outputParentDirectories` indexed by the path relative to the working directory), but only adjusted this call to `createOutputParentDirectory()` to use the joined path? Wouldn't that work?

pkg/builder/local_build_executor.go
cphang99: Won't that mean that if an outputDir is specified as `../../../foo` and workingDirectory is `bar` then the output will be `bar/../../../foo`?

Because of the way that directory traversal works in the `filesystem` package, then we'll fall foul of https://github.com/buildbarn/bb-storage/blob/master/pkg/filesystem/local_directory.go#L21 ?

I see the value of keeping that validation, so there's no possibility of entering a directory above the build directory.

EdSchouten: `path.Join()` already does a `path.Clean()` for you. this means that in your example, the resulting path will be `../../foo`. See: https://play.golang.org/p/2rjqZOac11q

There is no need to have any additional logic, as `createOutputParentDirectory()` will still detect build directory escapes.

pkg/builder/local_build_executor.go
```diff
@@ -240,8 +271,12 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 	// Build rules generally expect the parent directories to already be
 	// there. We later use the directory handles to extract output files.
 	outputParentDirectories := map[string]filesystem.Directory{}
+	workingDirectory := command.WorkingDirectory
+	if workingDirectory == "" {
```
EdSchouten: I just tested: there is likely no need to set `workingDirectory` in case it's empty. `path.Join()` does the right thing when the first argument is the empty string.

pkg/builder/local_build_executor.go
EdSchouten: So to clarify: revert tall of the changes to this source file, but replace this specific line with:
```go
dir, err := be.createOutputParentDirectory(buildDirectory, path.Join(command.WorkingDirectory, dirPath))
```

pkg/builder/local_build_executor.go
cphang99: Good point, that's what `path.Clean()` gives you. Thanks.

pkg/builder/local_build_executor.go
cphang99: Occam's razor for the win. Thanks @EdSchouten

PR #36, opened 2020-02-23
Switch to github actions for CI

cphang99:
This uses github inbuilt infrastructure in order to perform
CI.

Currently, will run gofmt, buildifier and run the
bb-remote-execution test suite


cphang99:
I'll post the log from my fork when it is done.

cphang99:
https://github.com/cphang99/bb-remote-execution/runs/463091211?check_suite_focus=true

Screenshot from 2020-02-23 13-37-14

EdSchouten:
Nice!

Something I should have brought up in your previous PR: would it be possible to call Gazelle instead of Buildifier here? Gazelle automatically does a Buildifier, so there is no need to do both. That said, let’s already go ahead and merge this.

PR #41, opened 2020-03-06
Fix creation of output directories ending with `..`

sdclarke:
After #35 paths containing `..` are now handled as defined by the REAPI spec, however a corner case has come up where a directory ending in `..` would end up being created and uploaded incorrectly due to the fact the determining the parent directory was done with a call to `path.Dir`. Extra logic has been added to deal with these cases in this pull request.

Given that this is related to the changes made in #35 I have added testing for this behaviour to the test written in that pull request.

### DIFF
--- a/pkg/builder/local_build_executor.go
+++ b/pkg/builder/local_build_executor.go
@@ -241,19 +241,6 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 	// there. We later use the directory handles to extract output files.
 	outputParentDirectories := map[string]filesystem.Directory{}
 	for _, outputDirectory := range command.OutputDirectories {
-		dirPath := path.Dir(outputDirectory)
-		if _, ok := outputParentDirectories[dirPath]; !ok {
-			dir, err := be.createOutputParentDirectory(buildDirectory, path.Join(command.WorkingDirectory, dirPath))
-			if err != nil {
-				attachErrorToExecuteResponse(response, err)
-				return response
-			}
-			outputParentDirectories[dirPath] = dir
-			if dir != buildDirectory {
-				defer dir.Close()
-			}
-		}
-
 		// Although REv2 explicitly document that only parents
 		// of output directories are created (i.e., not the
 		// output directory itself), Bazel recently changed its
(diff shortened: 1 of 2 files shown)


pkg/builder/local_build_executor.go
```diff
@@ -241,7 +241,12 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 	// there. We later use the directory handles to extract output files.
 	outputParentDirectories := map[string]filesystem.Directory{}
 	for _, outputDirectory := range command.OutputDirectories {
-		dirPath := path.Dir(outputDirectory)
+		var dirPath string
```
EdSchouten: Could you move this part into a helper function?

pkg/builder/local_build_executor_test.go
```diff
@@ -583,15 +583,23 @@ func TestLocalBuildExecutorWithWorkingDirectorySuccess(t *testing.T) {
 	// Creation of build/objects
 	rootDirectory := mock.NewMockDirectory(ctrl)
 	rootDirectory.EXPECT().Mkdir("build", os.FileMode(0777)).Return(nil)
+	rootDirectory.EXPECT().Mkdir("build", os.FileMode(0777)).Return(nil)
```
EdSchouten: Shouldn't this in the second case return `syscall.EEXIST`?

pkg/builder/local_build_executor_test.go
```diff
@@ -583,15 +583,23 @@ func TestLocalBuildExecutorWithWorkingDirectorySuccess(t *testing.T) {
...
 	buildDirectory := mock.NewMockDirectory(ctrl)
 	rootDirectory.EXPECT().Enter("build").Return(buildDirectory, nil)
+	rootDirectory.EXPECT().Enter("build").Return(buildDirectory, nil)
```
EdSchouten: Could you please keep this ordered a bit closer to what happens in practice? Also: maybe it's a bit more accurate to use two separate mocks for `buildDirectory` instead of one. In practice, it won't really happen that `Enter()` returns the same handle for both `Enter()` calls.

pkg/builder/local_build_executor_test.go
```diff
@@ -691,6 +718,9 @@ func TestLocalBuildExecutorWithWorkingDirectorySuccess(t *testing.T) {
 					"../hello.pic.d",
 					"objects/hello.pic.o",
 				},
+				OutputDirectories: []string{
```
EdSchouten: I'm not sure that this is a realistic example for a test case. Why would a build action ever need to contain an output path of that shape? Wouldn't that be the same as requesting `"."` (which also happens to be an interesting test case).

sdclarke: I chose this because it meant that all of the directories that I would need to know about in the test were ones that already existed as mocks. I'm happy to spend a bit longer creating something more realistic now that I have a bit of feedback though.

pkg/builder/local_build_executor_test.go
sdclarke: As part of making this test a bit more realistic I'll probably move away from needing to call `Enter("build")` twice so that should handle this issue.

pkg/builder/local_build_executor_test.go
sdclarke: As it turns out, requesting the output directory `"."` doesn't work currently either, I'll add a fix for that to this PR too.

sdclarke: > As it turns out, requesting the output directory `"."` doesn't work currently either, I'll add a fix for that to this PR too.

Latest update solves this.

sdclarke:
> Note that I have only compile tested this change. The unit tests no longer pass. Could you give this a spin, and if it works integrate it into your branch? Thanks!

I've integrated this and it still seems to work, I also fixed the failing unit tests.

PR #44, opened 2020-03-24
Sort worker platform properties

Qinusty:
Currently bb-worker expects the platform properties to be sorted by name as specified by the REAPI spec. Unfortunately this is quite inconvenient for the user configuring the workers as it does not cause the worker process to exit, it is constantly rejected by the scheduler for this. (Perhaps the scheduler needs to raise a different error which causes the worker to exit).

This PR sorts these before they are passed to the build client ensuring that any ordering of platform properties in the bb-worker configuration will be sorted by bb_worker to ensure that the spec is followed.

### DIFF
--- a/pkg/util/platforms.go
+++ b/pkg/util/platforms.go
@@ -1,0 +1,18 @@
+package util
+
+import (
+	"sort"
+
+	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+)
+
+type ByPlatformProperty []*remoteexecution.Platform_Property
+
+func (a ByPlatformProperty) Len() int           { return len(a) }
+func (a ByPlatformProperty) Less(i, j int) bool { return a[i].Name < a[j].Name }
+func (a ByPlatformProperty) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+
+func SortPlatformProperties(platform *remoteexecution.Platform) *remoteexecution.Platform {
+	sort.Sort(ByPlatformProperty(platform.Properties))
+	return platform
+}
--- a/pkg/util/platforms_test.go
+++ b/pkg/util/platforms_test.go
@@ -1,0 +1,23 @@
+package util_test
+
+import (
+	"fmt"
+	"testing"
+
+	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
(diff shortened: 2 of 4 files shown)


pkg/util/platforms_test.go
```diff
@@ -0,0 +1,23 @@
...
+func TestSortPlatformProperties(t *testing.T) {
+	reverseSortedPlatform := &remoteexecution.Platform{
+		Properties: []*remoteexecution.Platform_Property{
+			{Name: "os", Value: "armv6"},
+			{Name: "cpu", Value: "linux"},
+		},
+	}
+
+	sortedPlatform := re_util.SortPlatformProperties(reverseSortedPlatform)
+	if sortedPlatform.Properties[0].Name == "os" {
```
EdSchouten: Please use the `require` package to make this consistent with the rest of Buildbarn's unit tests.

Also, be sure to compare the entire resulting list. You should be able to write this:
```go
require.Equal(
   t,
   []*remoteexecution.Platform_Property{
          ...,
   },
   re_util.SortPlatformProperties([]*remoteexecution.Platform_Property{
          ...,
   }))
```

pkg/util/platforms.go
```diff
@@ -0,0 +1,18 @@
...
+	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+)
+
+type ByPlatformProperty []*remoteexecution.Platform_Property
+
+func (a ByPlatformProperty) Len() int           { return len(a) }
+func (a ByPlatformProperty) Less(i, j int) bool { return a[i].Name < a[j].Name }
+func (a ByPlatformProperty) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+
+func SortPlatformProperties(platform *remoteexecution.Platform) *remoteexecution.Platform {
```
EdSchouten: This function could just take the `[]*remoteexecution.Platform_Property` slice instead of the whole Platform message. That would make this function more reusable.

pkg/util/platforms_test.go
```diff
@@ -0,0 +1,23 @@
...
+		t.Errorf("Platform properties did not sort correctly")
+	}
+}
```
EdSchouten: Don't forget to run `gofmt` on your source files. That should fix CI for you.

EdSchouten:
Closing this due to inactivity.

PR #45, opened 2020-04-06
bb_runner chroot

sdclarke:
This allows bb_runner to chroot into the input root of the action it is running. This can be enabled by the ChrootIntoInputRoot boolean flag in the bb_runner configuration.

It also adds a second container image based on the alpine distribution which can be used when the chroot option is enabled as it is much more lightweight.

### DIFF
--- a/cmd/bb_runner/BUILD.bazel
+++ b/cmd/bb_runner/BUILD.bazel
@@ -1,5 +1,6 @@
 load("@com_github_buildbarn_bb_storage//tools:container.bzl", "container_push_official")
 load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_layer")
+load("@io_bazel_rules_docker//go:image.bzl", "go_image")
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
 go_library(
@@ -83,6 +84,13 @@ container_image(
     visibility = ["//visibility:public"],
 )
 
+go_image(
+    name = "bb_runner_bare_container",
+    embed = [":go_default_library"],
+    pure = "on",
+    visibility = ["//visibility:public"],
+)
+
 container_push_official(
     name = "bb_runner_ubuntu16_04_container_push",
     component = "bb-runner-ubuntu16-04",
@@ -93,4 +101,10 @@ container_push_official(
     name = "bb_runner_installer_push",
     component = "bb-runner-installer",
     image = ":bb_runner_installer",
+)
+
+container_push_official(
+    name = "bb_runner_bare_container_push",
+    component = "bb-runner-bare",
+    image = ":bb_runner_bare_container",
 )
--- a/cmd/bb_runner/main.go
+++ b/cmd/bb_runner/main.go
@@ -39,7 +39,8 @@ func main() {
 	r := runner.NewLocalRunner(
 		buildDirectory,
 		configuration.BuildDirectoryPath,
-		configuration.SetTmpdirEnvironmentVariable)
+		configuration.SetTmpdirEnvironmentVariable,
+		configuration.ChrootIntoInputRoot)
 
 	// When temporary directories need cleaning prior to executing a build
 	// action, attach a series of TemporaryDirectoryCleaningRunners.
--- a/pkg/proto/configuration/bb_runner/bb_runner.proto
+++ b/pkg/proto/configuration/bb_runner/bb_runner.proto
@@ -32,6 +32,8 @@ message ApplicationConfiguration {
   buildbarn.configuration.grpc.ClientConfiguration
       temporary_directory_installer = 6;
 
-  // Chroot into the input root to run commands
+  // Chroot into the input root to run commands. This option can be used
+  // if the input root contains a full userland installation. This
+  // feature is used by the BuildStream build system.
   bool chroot_into_input_root = 7;
 }
--- a/pkg/proto/configuration/bb_runner/bb_runner.proto
+++ b/pkg/proto/configuration/bb_runner/bb_runner.proto
@@ -31,4 +31,7 @@ message ApplicationConfiguration {
   // Optional helper process for resolving /tmp.
   buildbarn.configuration.grpc.ClientConfiguration
       temporary_directory_installer = 6;
+
+  // Chroot into the input root to run commands
+  bool chroot_into_input_root = 7;
 }
--- a/pkg/runner/local_runner.go
+++ b/pkg/runner/local_runner.go
@@ -24,15 +24,17 @@ type localRunner struct {
 	buildDirectory               filesystem.Directory
 	buildDirectoryPath           string
 	setTmpdirEnvironmentVariable bool
+	chrootIntoInputRoot          bool
 }
 
 // NewLocalRunner returns a Runner capable of running commands on the
 // local system directly.
-func NewLocalRunner(buildDirectory filesystem.Directory, buildDirectoryPath string, setTmpdirEnvironmentVariable bool) Runner {
+func NewLocalRunner(buildDirectory filesystem.Directory, buildDirectoryPath string, setTmpdirEnvironmentVariable bool, chrootIntoInputRoot bool) Runner {
 	return &localRunner{
 		buildDirectory:               buildDirectory,
 		buildDirectoryPath:           buildDirectoryPath,
 		setTmpdirEnvironmentVariable: setTmpdirEnvironmentVariable,
+		chrootIntoInputRoot:          chrootIntoInputRoot,
 	}
 }
 
@@ -70,10 +72,18 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
 	if len(request.Arguments) < 1 {
 		return nil, status.Error(codes.InvalidArgument, "Insufficient number of command arguments")
 	}
-	cmd := exec.CommandContext(ctx, request.Arguments[0], request.Arguments[1:]...)
-	// TODO: Convert WorkingDirectory and TemporaryDirectory to use
-	// platform specific path delimiters.
-	cmd.Dir = filepath.Join(r.buildDirectoryPath, request.InputRootDirectory, request.WorkingDirectory)
+	var cmd *exec.Cmd
+	if r.chrootIntoInputRoot {
+		envPrependedArguments := []string{"/usr/bin/env", "--"}
+		envPrependedArguments = append(envPrependedArguments, request.Arguments...)
+		cmd = exec.CommandContext(ctx, envPrependedArguments[0], envPrependedArguments[1:]...)
+		// TODO: Convert WorkingDirectory and TemporaryDirectory to use
+		// platform specific path delimiters.
+		cmd.Dir = filepath.Join("/", request.WorkingDirectory)
+	} else {
+		cmd = exec.CommandContext(ctx, request.Arguments[0], request.Arguments[1:]...)
+		cmd.Dir = filepath.Join(r.buildDirectoryPath, request.InputRootDirectory, request.WorkingDirectory)
+	}
 	cmd.Env = make([]string, 0, len(request.EnvironmentVariables)+1)
 	if r.setTmpdirEnvironmentVariable && request.TemporaryDirectory != "" {
 		cmd.Env = append(cmd.Env, "TMPDIR="+filepath.Join(r.buildDirectoryPath, request.TemporaryDirectory))
@@ -95,7 +105,14 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
 		return nil, util.StatusWrap(err, "Failed to open stderr")
 	}
 	cmd.Stderr = stderr
-	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	if r.chrootIntoInputRoot {
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Chroot:  path.Join(r.buildDirectoryPath, request.InputRootDirectory),
+			Setpgid: true,
+		}
+	} else {
+		cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	}
 	// Start the subprocess. We can already close the output files
 	// while the process is running.
 	err = cmd.Start()
--- a/pkg/runner/local_runner.go
+++ b/pkg/runner/local_runner.go
@@ -74,6 +74,11 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
 	}
 	var cmd *exec.Cmd
 	if r.chrootIntoInputRoot {
+		// The addition of /usr/bin/env is necessary as the PATH resolution
+		// will take place prior to the chroot, so the executable may not be
+		// found by exec.LookPath() inside exec.CommandContext() and may
+		// cause cmd.Start() to fail when it shouldn't.
+		// https://github.com/golang/go/issues/39341
 		envPrependedArguments := []string{"/usr/bin/env", "--"}
 		envPrependedArguments = append(envPrependedArguments, request.Arguments...)
 		cmd = exec.CommandContext(ctx, envPrependedArguments[0], envPrependedArguments[1:]...)
--- a/pkg/runner/local_runner_test.go
+++ b/pkg/runner/local_runner_test.go
@@ -31,7 +31,7 @@ func TestLocalRunner(t *testing.T) {
 		// variables should cause the process to be executed in
 		// an empty environment. It should not inherit the
 		// environment of the runner.
-		runner := runner.NewLocalRunner(buildDirectory, buildDirectoryPath, false)
+		runner := runner.NewLocalRunner(buildDirectory, buildDirectoryPath, false, false)
 		response, err := runner.Run(context.Background(), &runner_pb.RunRequest{
 			Arguments:          []string{"/usr/bin/env"},
 			StdoutPath:         "EmptyEnvironment/stdout",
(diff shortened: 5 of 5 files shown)


cmd/bb_runner/BUILD.bazel
```diff
@@ -65,6 +65,25 @@ container_image(
     visibility = ["//visibility:public"],
 )
 
+container_image(
```
EdSchouten, line 68: Do you actually need anything from the Alpine container image? Wouldn't it be sufficient to use a plain `go_image()` for this bare use case, just like we already do for bb_scheduler and bb_worker?

If you can't use `go_image()` literally, you can at least consider using `base = "@go_image_static//image"`, which we also use for bb_browser.

cmd/bb_runner/BUILD.bazel
```diff
@@ -65,6 +65,25 @@ container_image(
...
+    entrypoint = [
+        "/tini",
+        "-v",
+        "-g",
+        "--",
+        "/bb_runner",
+    ],
+    layers = [
+        ":bb_runner_layer",
+        ":passwd_layer",
```
EdSchouten, line 80: I think you don't need a custom passwd file for this container image. It was only there to appease build actions that try to resolve their home directory or username. This isn't relevant if actions end up chrooting, as they will not be able to access this passwd file.

cmd/bb_runner/BUILD.bazel
```diff
@@ -65,6 +65,25 @@ container_image(
...
+        ":tini_layer",
+    ],
+    user = "root",
```
EdSchouten, line 83: What can we do to keep this somewhat secure? Running build actions as root doesn't seem like a good idea. Can't we keep it running as a non-root user ID, but configure Docker/Kubernetes/.... to grant it `CAP_SYS_CHROOT`?

pkg/runner/local_runner.go
```diff
@@ -66,14 +68,60 @@ func convertTimeval(t syscall.Timeval) *duration.Duration {
 	}
 }
 
+func (r *localRunner) createDev(inputRootPath string) error {
```
EdSchouten, line 71: There are a couple of things I don't really like about this approach:
- The chroot option is currently conflated with character device creation, while you could argue the two are independent.
- The list of character devices is Linux specific and not flexible. On FreeBSD, /dev/urandom is a symbolic link.
- We don't use any of the `filesystem.Directory` API here, meaning that this code might be prone to symlink following attacks.
- It makes bb_runner partially responsible for setting up the file system, while that is originally what bb_worker is responsible for.
- Relatedly, it makes bb_runner fatter. The idea is that bb_runner remains relatively simple, so people can also write their own ones where needed. Adding the chroot support is all right, but I think that device node creation is unnecessary.

In other words, can't we somehow make this part of bb_worker? Maybe we should extend LocalBuildExecutor and/or BuildDirectory?

pkg/runner/local_runner.go
```diff
@@ -66,14 +68,60 @@ func convertTimeval(t syscall.Timeval) *duration.Duration {
...
+		mode := realdev.Mode &^ syscall.S_IRWXG
+		mode = mode | ((mode & syscall.S_IRWXO) << 3)
+		if err := syscall.Mknod(finalPath, mode, int(realdev.Rdev)); err != nil {
+			return util.StatusWrapf(err, "Mknod failed for %#v", finalPath)
+		}
+	}
+	return nil
+}
+
+func (r *localRunner) unlinkDev(inputRootPath string) error {
```
EdSchouten, line 96: This logic isn't needed. bb_worker will clean up the build directory for you.

EdSchouten:
I think we should consider limiting this PR to chroot support for bb_runner first. Adding support for device nodes is something we may add separately.

cmd/bb_runner/BUILD.bazel
EdSchouten: Also be sure to add a `container_push_official()`, so that we provide ready to use container images. How should we call this? `bb-runner-bare`?

sdclarke:
I've run into a problem when making these changes and I'm not sure how to get around it. Basically when using a bare `go_image()`, the `exec.CommandContext()` call will set the `lookPathErr` field of the command, as there is nothing (I assume) in /bin etc. This means that when the command comes to be run, it will fail with an error message like:
```
Failed to run command: Failed to start process: exec: "sh": executable file not found in $PATH
```
There may be some hacking I can do involving setting the `PATH` environment variable but this seems complex and strikes me as a bas idea.

I've managed to hack my way around it by joining `/bin` onto `request.Arguments[0]` in `CommandContext`, but obviously this doesn't work if the command isn't in `/bin`. This did allow me to confirm that if I can get around this problem everything should be working however.

pkg/runner/local_runner.go
```diff
@@ -66,14 +69,60 @@ func convertTimeval(t syscall.Timeval) *duration.Duration {
...
+}
+
+func (r *localRunner) lookPathChroot(file, chroot string) (string, error) {
+	if strings.Contains(file, "/") {
+		if err := findExecutable(file); err != nil {
+			return file, err
+		}
+		return file, nil
+	}
+	path := os.Getenv("PATH")
```
EdSchouten, line 90: I don't think this should use `os.Getenv()` to obtain `$PATH`. Instead, it should extract it from the environment stored in `RunRequest`.

The nice thing about a feature like this is that it can easily be implemented as a new decorator type for `Runner`. Think of it as a simple mutator for requests that come in: You get a `RunRequest` containing an `argv[0]` without a slash in it. You then try to resolve it to one that does, based on the `PATH` specified in its environment.

By modelling it that way, it becomes a lot easier to unit test this functionality.

pkg/runner/chroot_runner.go
```diff
@@ -0,0 +1,73 @@
...
+
+type chrootRunner struct {
+	base               Runner
+	buildDirectoryPath string
+}
+
+// NewChrootRunner modifies the arguments of the incoming RunRequest so
+// that the first argument is an absolute path resolved using the path
+// in the request using the input root as the chroot
+func NewChrootRunner(base Runner, buildDirectoryPath string) Runner {
```
EdSchouten, line 24: Maybe it would be better to call this Argv0ResolvingRunner, or ExecutablePathResolvingRunner?

Could it be made to work for non-chrooted requests as well? That way we could simply enable it unconditionally, thereby getting consistent behaviour between chroot and !chroot.

pkg/runner/chroot_runner.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	if m := d.Mode(); !m.IsDir() && m&0111 != 0 {
+		return nil
+	}
+	return os.ErrPermission
+}
+
+func (r *chrootRunner) lookPathChroot(request *runner.RunRequest) (string, error) {
+	file := request.Arguments[0]
+	if strings.Contains(file, "/") {
```
EdSchouten, line 44: ContainsRune() ?

pkg/runner/chroot_runner.go
```diff
@@ -0,0 +1,73 @@
...
+		return err
+	}
+	if m := d.Mode(); !m.IsDir() && m&0111 != 0 {
+		return nil
+	}
+	return os.ErrPermission
+}
+
+func (r *chrootRunner) lookPathChroot(request *runner.RunRequest) (string, error) {
+	file := request.Arguments[0]
```
EdSchouten, line 43: The request could also, unfortunately, contain zero arguments.

pkg/runner/chroot_runner.go
```diff
@@ -0,0 +1,73 @@
...
+	if strings.Contains(file, "/") {
+		if err := findExecutable(file); err != nil {
+			return file, err
+		}
+		return file, nil
+	}
+	for _, dir := range filepath.SplitList(request.EnvironmentVariables["PATH"]) {
```
EdSchouten, line 50: Inside Run() you already did a map lookup. Maybe nicer to save that value and pass it in?

pkg/runner/chroot_runner.go
```diff
@@ -0,0 +1,73 @@
...
+		if dir == "" {
```
EdSchouten, line 51: I think `filepath.Join()` already does the right thing when the string is empty.

pkg/runner/chroot_runner.go
```diff
@@ -0,0 +1,73 @@
+package runner
```
EdSchouten, line 1: Is there a way we could get unit testing coverage for this type?

Using “pkg/filesystem” instead of “os” might be helpful to achieve that.

pkg/runner/executable_path_resolving_runner.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/buildbarn/bb-remote-execution/pkg/proto/runner"
+	"github.com/buildbarn/bb-storage/pkg/filesystem"
+	"github.com/buildbarn/bb-storage/pkg/util"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type rootDirectoryOpener func(inputRootDirectory string) (filesystem.DirectoryCloser, error)
```
EdSchouten, line 16: Rule of thumb: don't let a public API depend on private types. For example, don't let public functions take private types or return them. This makes them less flexible.

By making it public, you could add code to internal/mock to automatically generate a mock for RootDirectoryOpener, thereby making the unit tests a bit cleaner.

pkg/runner/executable_path_resolving_runner.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	d, err := dir.Lstat(filepath.Base(file))
+	if err != nil {
+		return err
+	}
+	if m := d.Type(); m == filesystem.FileTypeExecutableFile {
+		return nil
+	} else if m == filesystem.FileTypeSymlink {
+		for i := 0; i < 10; i++ {
+			link, err := dir.Readlink(filepath.Base(file))
```
EdSchouten, line 65: What if you moved the top part of this loop until the Lstat() call to the bottom of the loop? I suspect that would allow you to remove some of the code above the loop?

A bit more explanation what I mean: Right now this entire function is somewhat like this:
```go
a
for {
  b
  a
}
```
But maybe it could be translated to:
```go
for {
  a
  b
}
```

pkg/runner/executable_path_resolving_runner.go
```diff
@@ -0,0 +1,121 @@
...
+	return file, status.Error(codes.InvalidArgument, "Executable file not found in PATH")
+}
+
+func (r *executablePathResolvingRunner) Run(ctx context.Context, request *runner.RunRequest) (*runner.RunResponse, error) {
+	if len(request.Arguments) < 1 {
+		return nil, status.Error(codes.InvalidArgument, "Insufficient number of command arguments")
+	}
+	path, present := request.EnvironmentVariables["PATH"]
+	if !present {
+		return nil, status.Error(codes.InvalidArgument, "No PATH in command's environment variables")
```
EdSchouten, line 113: This seems a bit too strict, right? It means that if a build action provides an `argv[0]` with an absolute path (or at least one containing a slash), it must still provide `PATH`.

cmd/bb_runner/main.go
```diff
@@ -36,10 +38,38 @@ func main() {
 		log.Fatal("Failed to open build directory: ", err)
 	}
 
-	r := runner.NewLocalRunner(
-		buildDirectory,
-		configuration.BuildDirectoryPath,
-		configuration.SetTmpdirEnvironmentVariable)
+	rootDirectoryOpener := func(inputRootDirectory string) (filesystem.DirectoryCloser, error) {
```
EdSchouten, line 41: It may be wiser to convert this to:
```go
var rootDirectoryOpener runner.RootDirectoryOpener
if configuration.ChrootIntoInputRoot {
   rootDirectoryOpener = func(...) {
       ...
   }
} else {
   rootDirectoryOpener = func(...) {
       ...
   }
}
```
That way the boolean comparison is only performed once during startup.

Another thing that you could consider doing is opening the root directory just once during bb_runner startup. You then need to make sure that this directory is never closed, but you can solve that by calling `filesystem.NopDirectoryCloser()` to give a `filesystem.Directory` a no-op `Close()` operation.

pkg/runner/executable_path_resolving_runner_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	inputRootDirectory := mock.NewMockDirectoryCloser(ctrl)
+
+	mockRootDirectoryOpener := func(inputRootPath string) (filesystem.DirectoryCloser, error) {
+		return inputRootDirectory, nil
+	}
+
+	executablePathResolvingRunner := runner.NewExecutablePathResolvingRunner(baseRunner, mockRootDirectoryOpener)
+
+	inputRootBin := mock.NewMockDirectoryCloser(ctrl)
```
EdSchouten, line 127:
```go
inputRootBin.EXPECT().Close()
```
?

pkg/runner/executable_path_resolving_runner_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"context"
+	"testing"
+
+	"github.com/buildbarn/bb-remote-execution/internal/mock"
+	runner_pb "github.com/buildbarn/bb-remote-execution/pkg/proto/runner"
+	"github.com/buildbarn/bb-remote-execution/pkg/runner"
+	"github.com/buildbarn/bb-storage/pkg/filesystem"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
```
EdSchouten, line 13: Please separate imports by second level domain name: just add a newline between all the github.com and golang.org.

pkg/runner/executable_path_resolving_runner.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (r *executablePathResolvingRunner) enterDir(file, chroot string) (filesystem.DirectoryCloser, error) {
+	rootDir, err := r.rootDirectoryOpener(chroot)
+	if err != nil {
+		return nil, util.StatusWrap(err, "Failed to open root directory")
+	}
+	components := strings.FieldsFunc(file, func(r rune) bool { return r == '/' })
+	dir := rootDir
+	for n, component := range components[:len(components)-1] {
+		dir2, err := dir.EnterDirectory(component)
```
EdSchouten, line 42: I think this doesn't work when a directory is a symlink, right? For example, consider the case where:
```
PATH=/bin
```
And `/bin` is a symlink to `usr/bin`.

sdclarke: Would it be acceptable in the case you described to have the eventual value of `Arguments[0]` still be `/bin/<executable>`. A very quick test suggests that `exec.Command` will still work but there may be an edge case I'm not aware of.

pkg/runner/local_runner.go
```diff
@@ -73,7 +75,11 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
 	cmd := exec.CommandContext(ctx, request.Arguments[0], request.Arguments[1:]...)
 	// TODO: Convert WorkingDirectory and TemporaryDirectory to use
 	// platform specific path delimiters.
-	cmd.Dir = filepath.Join(r.buildDirectoryPath, request.InputRootDirectory, request.WorkingDirectory)
+	if r.chrootIntoInputRoot {
+		cmd.Dir = filepath.Join("/", request.WorkingDirectory)
```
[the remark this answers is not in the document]
sdclarke: Unless it's implicit or I've missed it looking through the code, I can't see where a Chdir would happen. Do you know for sure that it does?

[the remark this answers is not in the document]
EdSchouten: Ah, you're correct. I just checked go's src/syscall/exec_linux.go and it's pretty bare.

Calling chdir("/") after chroot() is one of those typical UNIX best practices that you typically see being used, for the reason that without doing that there are easy ways to escape. It looks like Go sort of assumes the caller tidies everything up beforehand.

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -31,4 +31,7 @@ message ApplicationConfiguration {
   // Optional helper process for resolving /tmp.
   buildbarn.configuration.grpc.ClientConfiguration
       temporary_directory_installer = 6;
+
+  // Chroot into the input root to run commands
```
EdSchouten, line 35: Could you extend this comment to explain why/when you'd want this? Also a perfect moment to promote Buildstream!

"This option can be used if the input root contains a full userland installation. This feature is used by the BuildStream build system."

pkg/runner/local_runner.go
```diff
@@ -70,10 +72,18 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
 	if len(request.Arguments) < 1 {
 		return nil, status.Error(codes.InvalidArgument, "Insufficient number of command arguments")
 	}
-	cmd := exec.CommandContext(ctx, request.Arguments[0], request.Arguments[1:]...)
-	// TODO: Convert WorkingDirectory and TemporaryDirectory to use
-	// platform specific path delimiters.
-	cmd.Dir = filepath.Join(r.buildDirectoryPath, request.InputRootDirectory, request.WorkingDirectory)
+	var cmd *exec.Cmd
+	if r.chrootIntoInputRoot {
```
EdSchouten, line 76: Could you add a comment here to explain why the `/usr/bin/env` is added? If you feel like filing a bug against upstream Go, please do! Feel free to include a link to that bug report in the comment.

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -32,6 +32,8 @@ message ApplicationConfiguration {
   buildbarn.configuration.grpc.ClientConfiguration
       temporary_directory_installer = 6;
 
-  // Chroot into the input root to run commands
+  // Chroot into the input root to run commands This option can be used
```
EdSchouten, line 35: Missing period between the sentences. ;-)

pkg/runner/local_runner.go
```diff
@@ -74,6 +74,11 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
 	}
 	var cmd *exec.Cmd
 	if r.chrootIntoInputRoot {
+		// The addition of /usr/bin/env is necessary as the runner will run
+		// in a bare container, so the executable won't be found by
```
EdSchouten, line 78: Nit: This isn't necessarily due to the fact that bb-runner is run in a bare container. It's because PATH resolution is done prior to chroot.

PR #48, opened 2020-04-09
Avoid caching directory filehandles during action execution (fixes #46)

### LINKED ISSUE #46: Action execution fails if worker-created output directory is deleted
This appears to have been introduced by #41 (testing without that commit makes the problem go away). The output directory is now created by the worker, but filesystem.Directory follows the fd instead of the path, so the worker will only try to upload the precise directory it created.

If the action does something like this:

```
rmdir out
mkdir out
```

Then the action will fail, reporting that the output directory could not be found.

Notably, @io_bazel_rules_go//:stdlib does this.

asuffield:
Refactor this code so that we can run the same logic before and after
execution: once to create the output directories, then again to find
them by path. This lets us handle the case where the action deletes
and recreates the directory.

### DIFF
--- a/pkg/builder/local_build_executor.go
+++ b/pkg/builder/local_build_executor.go
@@ -131,9 +131,9 @@ func (be *localBuildExecutor) uploadTree(ctx context.Context, outputDirectory fi
 	return treeDigest, err
 }
 
-func getOutputDirectory(inputRootDirectory filesystem.Directory, outputPath string, create bool) (filesystem.DirectoryCloser, error) {
+func getDirectory(inputRootDirectory filesystem.Directory, outputParentPath string, create bool) (filesystem.DirectoryCloser, error) {
 	// Create and enter successive components, closing the former.
-	components := strings.FieldsFunc(outputPath, func(r rune) bool { return r == '/' })
+	components := strings.FieldsFunc(outputParentPath, func(r rune) bool { return r == '/' })
 	d := filesystem.NopDirectoryCloser(inputRootDirectory)
 	for n, component := range components {
 		if component != "." {
@@ -146,7 +146,7 @@ func getOutputDirectory(inputRootDirectory filesystem.Directory, outputPath stri
 			d2, err := d.EnterDirectory(component)
 			d.Close()
 			if err != nil {
-				return nil, util.StatusWrapf(err, "Failed to enter output directory %#v", path.Join(components[:n+1]...))
+				return nil, util.StatusWrapf(err, "Failed to enter directory %#v", path.Join(components[:n+1]...))
 			}
 			d = d2
 		}
@@ -154,16 +154,8 @@ func getOutputDirectory(inputRootDirectory filesystem.Directory, outputPath stri
 	return d, nil
 }
 
-type outputDirectories map[string]filesystem.DirectoryCloser
-
-func (d outputDirectories) Close() {
-	for _, dir := range d {
-		dir.Close()
-	}
-}
-
-func newOutputDirectories(command *remoteexecution.Command, inputRootDirectory filesystem.Directory, create bool) (outputDirectories, error) {
-	paths := map[string]struct{}{}
+func outputDirectories(command *remoteexecution.Command) []string {
+	pathMap := map[string]struct{}{}
 
 	for _, outputDirectory := range command.OutputDirectories {
 		// Although REv2 explicitly document that only parents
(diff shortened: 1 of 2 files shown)


pkg/builder/local_build_executor.go
```diff
@@ -356,6 +382,13 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 		response.Result.StderrDigest = stderrDigest.GetPartialDigest()
 	}
 
+	outputParentDirectories, err := newOutputDirectories(command, inputRootDirectory, false)
```
EdSchouten, line 385: This is kind of odd, as it ends up doing `Mkdir()` calls during the upload phase...? I suspect we simply shouldn't share the logic.

pkg/builder/local_build_executor.go
```diff
@@ -356,6 +382,13 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
...
+	if err != nil {
+		attachErrorToExecuteResponse(response, err)
+		return response
+	}
+	defer outputParentDirectories.Close()
+
 	// Upload output files.
 	for _, outputFile := range command.OutputFiles {
 		outputParentDirectory := outputParentDirectories[path.Dir(outputFile)]
```
EdSchouten, line 394: If we're going down this road, we should likely partially revert 855598d6fec55bfa0d15f5d1d58d55a061b04b4e. It may now well be the case that output directories no longer exist or have become symlinks. Those cases now also need to be handled.

pkg/builder/local_build_executor.go
asuffield: Note that the third parameter suppresses the Mkdir - everything else is identical

EdSchouten: Ah, gotcha.

allada:
Oh wow, I guess I should have checked the pull requests before I fixed this.
I believe I am fixing the same thing here:
https://github.com/buildbarn/bb-remote-execution/pull/49

pkg/builder/local_build_executor.go
asuffield: I think I have now untangled this.

asuffield:
Unit tests are updated to match the code change, although I am unhappy with the level of mockery here.

pkg/builder/local_build_executor_test.go
```diff
@@ -325,13 +329,14 @@ func TestLocalBuildExecutorOutputSymlinkReadingFailure(t *testing.T) {
...
 	buildDirectory.EXPECT().InstallHooks(filePool)
 	buildDirectory.EXPECT().Mkdir("root", os.FileMode(0777))
-	inputRootDirectory := mock.NewMockBuildDirectory(ctrl)
 	buildDirectory.EXPECT().EnterBuildDirectory("root").Return(inputRootDirectory, nil)
 	inputRootDirectory.EXPECT().MergeDirectoryContents(
 		ctx,
 		digest.MustNewDigest("nintendo64", "7777777777777777777777777777777777777777777777777777777777777777", 42),
 	).Return(nil)
 	inputRootDirectory.EXPECT().Mkdir("foo", os.FileMode(0777)).Return(nil)
+	inputRootDirectory.EXPECT().EnterDirectory("foo").Return(fooDirectory, nil)
```
EdSchouten, line 338: These calls don't seem necessary, right?

pkg/builder/local_build_executor_test.go
```diff
@@ -646,12 +667,44 @@ func TestLocalBuildExecutorWithWorkingDirectorySuccess(t *testing.T) {
...
+	objectsDirectory.EXPECT().Lstat("hello.pic.o").Return(
+		filesystem.NewFileInfo("hello.pic.o", filesystem.FileTypeExecutableFile),
+		nil)
+	objectsDirectory.EXPECT().Close()
 	contentAddressableStorage.EXPECT().PutFile(ctx, objectsDirectory, "hello.pic.o", gomock.Any()).Return(
 		digest.MustNewDigest("ubuntu1804", "0000000000000000000000000000000000000000000000000000000000000008", 890),
 		nil)
+
+	inputRootDirectory.EXPECT().EnterDirectory("outputParent").Return(outputParentDirectory0, nil)
+	outputParentDirectory0.EXPECT().Lstat("..").Return(
```
EdSchouten, line 698: You can't use the filesystem.Directory API to traverse upwards in the file system hierarchy.

pkg/builder/local_build_executor_test.go
asuffield: It's the createOutputDirectories function, which uses getDirectory just to create parents and then closes the result.

The alternative is duplicating most of getDirectory, which seems worse.

pkg/builder/local_build_executor_test.go
asuffield: Tracked down the missing path.Join before a path.Dir. Also cleaned up the error handling in this scenario and added a test for it specifically, because this isn't obvious and eludes the rest of the tests.

EdSchouten:
This should be fixed in https://github.com/buildbarn/bb-remote-execution/commit/55f422090559f6b03f9930538e975d2525b95688!

PR #49, opened 2020-04-12
Fix bug when action replaces folder we upload nothing

allada:
Fixes a bug where if an action deletes or moves an output
folder created by bb_worker, bb_worker will not upload any
results because we re-use the same file handle we used to
create the directory.


EdSchouten:
This should be fixed in https://github.com/buildbarn/bb-remote-execution/commit/55f422090559f6b03f9930538e975d2525b95688!

PR #50, opened 2020-04-13
Allows multiple files/folders to be downloaded at one time

allada:
Enables the ability to multiplex downloading of files/folders
by introducing SharedActionQueue utility that acts similar
to a promise/future, but more specialized for thread pools.

---
This change is [Reviewable](https://reviewable.io/reviews/buildbarn/bb-remote-execution/50)

### DIFF
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -21,6 +21,7 @@ import (
 	runner_pb "github.com/buildbarn/bb-remote-execution/pkg/proto/runner"
 	"github.com/buildbarn/bb-remote-execution/pkg/runner"
 	"github.com/buildbarn/bb-remote-execution/pkg/sync"
+	re_util "github.com/buildbarn/bb-remote-execution/pkg/util"
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
 	blobstore_configuration "github.com/buildbarn/bb-storage/pkg/blobstore/configuration"
 	"github.com/buildbarn/bb-storage/pkg/cas"
@@ -74,6 +75,11 @@ func main() {
 		}
 		filePool = re_filesystem.NewDirectoryBackedFilePool(filePoolDirectory)
 	}
+
+	if configuration.SharedActionQueueSize < 1 {
+		log.Fatal("shared_action_queue_size must be set and greater than zero.")
+	}
+	sharedActionQueue := re_util.NewSharedActionQueue(configuration.SharedActionQueueSize)
 
 	// Storage access.
 	contentAddressableStorageBlobAccess, actionCache, err := blobstore_configuration.CreateBlobAccessObjectsFromConfig(
@@ -203,7 +209,9 @@ func main() {
 				case *bb_worker.ApplicationConfiguration_LocalBuildDirectory:
 					buildDirectory = builder.NewNaiveBuildDirectory(
 						naiveBuildDirectory,
-						contentAddressableStorage)
+						contentAddressableStorage,
+						sharedActionQueue,
+					)
 				}
 
 				// Create a per-action subdirectory in
--- a/pkg/builder/naive_build_directory.go
+++ b/pkg/builder/naive_build_directory.go
@@ -5,6 +5,7 @@ import (
 	"path"
 
 	re_filesystem "github.com/buildbarn/bb-remote-execution/pkg/filesystem"
+	re_util "github.com/buildbarn/bb-remote-execution/pkg/util"
 	"github.com/buildbarn/bb-storage/pkg/cas"
 	"github.com/buildbarn/bb-storage/pkg/digest"
 	"github.com/buildbarn/bb-storage/pkg/filesystem"
@@ -14,6 +15,7 @@ import (
 type naiveBuildDirectory struct {
 	filesystem.DirectoryCloser
 	contentAddressableStorage cas.ContentAddressableStorage
+	sharedActionQueue         *re_util.SharedActionQueue
 }
 
 // NewNaiveBuildDirectory creates a BuildDirectory that is backed by a
@@ -26,10 +28,11 @@ type naiveBuildDirectory struct {
 // regular local file systems. The downside of such file systems is that
 // we cannot populate them on demand. All of the input files must be
 // present before invoking the build action.
-func NewNaiveBuildDirectory(directory filesystem.DirectoryCloser, contentAddressableStorage cas.ContentAddressableStorage) BuildDirectory {
+func NewNaiveBuildDirectory(directory filesystem.DirectoryCloser, contentAddressableStorage cas.ContentAddressableStorage, sharedActionQueue *re_util.SharedActionQueue) BuildDirectory {
 	return &naiveBuildDirectory{
 		DirectoryCloser:           directory,
 		contentAddressableStorage: contentAddressableStorage,
+		sharedActionQueue:         sharedActionQueue,
 	}
 }
 
@@ -41,6 +44,7 @@ func (d *naiveBuildDirectory) EnterBuildDirectory(name string) (BuildDirectory, 
 	return &naiveBuildDirectory{
 		DirectoryCloser:           child,
 		contentAddressableStorage: d.contentAddressableStorage,
+		sharedActionQueue:         d.sharedActionQueue,
 	}, nil
 }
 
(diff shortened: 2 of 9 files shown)


allada:
I'm thinking this utility might fit better in `bb_storage`, since I believe it also suffers from the same issues, but for now I'll keep it here and keep pushing PRs related to scaling issues.

cmd/bb_worker/main.go
```diff
@@ -75,6 +76,12 @@ func main() {
 		filePool = re_filesystem.NewDirectoryBackedFilePool(filePoolDirectory)
 	}
 
+	sharedActionQueueSize := configuration.SharedActionQueueSize
+	if sharedActionQueueSize == 0 {
+		sharedActionQueueSize = 255
```
EdSchouten, line 81: Please don't add defaults like these. Current Buildbarn style is to refrain from setting default size constants. This is done intentionally. The size of this queue would strongly depend on how people deploy Buildbarn. For very small workers, 255 would be far too high.

pkg/util/shared_action_queue.go
```diff
@@ -0,0 +1,93 @@
...
+type ActionGroup = []Action
+
+type SharedActionQueue struct {
+	sharedActionQueue chan func()
+}
+
+// This is a utility designed to run 'Action's to do some work async but limit
+// by a pre-defined number. This allows multiple components to share the same
+// pool of goroutines or other similar resources (like data streams) easily.
+func NewSharedActionQueue(maxConcurrentDownloads uint32) *SharedActionQueue {
```
EdSchouten, line 23: One important aspect of having queueing for this is that it respects `context.Context`. This queue doesn't seem to do that. This means that workers may get stuck on downloading files for certain build actions, even if the client that requested the build action to be run disappears.

cmd/bb_worker/main.go
allada: In that case would you like me to default it to one? We can't default it to zero as it will cause the queue to never run. Alternatively I can make it whatever this number is +1 to get around this.

EdSchouten: Sure, or call `log.Fatal()` in case it's `<= 0`, so that the user comes up with an explicit value for it.

allada: Done.

pkg/util/shared_action_queue.go
allada: Done. It now splits the context and up to the consumer to use `ActionCache.Ctx()`.

allada:
I'm confused on why buildifier wants me to delete this test, any hints would be useful.

pkg/util/shared_action_queue_test.go
```diff
@@ -0,0 +1,147 @@
+package sync_test
```
EdSchouten, line 1: I think you need to rename this to `util_test` to keep Gazelle happy.

pkg/builder/naive_build_directory_test.go
```diff
@@ -76,19 +77,20 @@ func TestNaiveBuildDirectorySuccess(t *testing.T) {
...
 	contentAddressableStorage.EXPECT().GetFile(
-		ctx,
+		gomock.AssignableToTypeOf(ctx),
 		digest.MustNewDigest("netbsd", "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", 512),
 		buildDirectory,
 		"executable",
 		true).Return(nil)
 	buildDirectory.EXPECT().Symlink("executable", "link-to-executable").Return(nil)
-	inputRootPopulator := builder.NewNaiveBuildDirectory(buildDirectory, contentAddressableStorage)
+	shared_action_queue := re_util.NewSharedActionQueue(10)
```
allada, line 92: todo: caps vs snake case

pkg/util/shared_action_queue_test.go
allada: Done.

allada: Done.

allada:
I'm still doing a bit of testing on this branch, so don't merge yet please.

pkg/util/shared_action_queue.go
```diff
@@ -0,0 +1,100 @@
...
```
EdSchouten, line 26: `s/maxConcurrentDownloads/maxConcurrency/`? This code is quite independent from doing downloads.

pkg/builder/naive_build_directory.go
```diff
@@ -56,36 +60,67 @@ func (d *naiveBuildDirectory) mergeDirectoryContents(ctx context.Context, digest
...
+			if err != nil {
+				actionContext.Done(util.StatusWrapf(err, "Failed to enter input directory %#v", path.Join(childComponents...)))
+				return
+			}
+			// We don't want to close the folder's handle until our children are done, however we also don't
+			// want to hold a slot of the SharedActionQueue (or we might deadlock). Since 'mergeDirectoryContents'
+			// is a blocking function, we can notify the ActionContext that we want to release our hold on a slot
+			// and let a new goroutine block on 'mergeDirectoryContents' for us then call the result handler
+			// when we have finished.
+			doneHandler := actionContext.Defer()
```
EdSchouten, line 107: To me it's not entirely clear what's the use of `actionContext.Defer()`. Couldn't you just call `actionContext.Done()` inside the goroutine below?

pkg/builder/naive_build_directory.go
```diff
@@ -56,36 +60,67 @@ func (d *naiveBuildDirectory) mergeDirectoryContents(ctx context.Context, digest
...
-		}
+		directoryName := directory.Name // Ensure we don't grab the reference to an object.
+		actionGroup = append(actionGroup, func(actionContext *re_util.ActionContext) {
+			childDirectory, err := inputDirectory.EnterDirectory(directoryName)
+			if err != nil {
+				actionContext.Done(util.StatusWrapf(err, "Failed to enter input directory %#v", path.Join(childComponents...)))
+				return
+			}
+			// We don't want to close the folder's handle until our children are done, however we also don't
+			// want to hold a slot of the SharedActionQueue (or we might deadlock). Since 'mergeDirectoryContents'
```
EdSchouten, line 103: The deadlock we would otherwise run into is in my opinion a symptom of us trying to introduce too much parallelism. Imagine this input root that has an awful number of directories. Millions and millions of directories, each nested into each other. Just ignore any of the files that are inside of them (if any). Without the defer logic in here, the queue would deadlock, for the reason that all of the slots are filled with directory processing actions, all trying to yield even more actions that process directories.

The defer logic/goroutine breakup you added prevents the deadlock, but the question is what we lose as a result of that. The answer to that is that there's no upper bound on the concurrent number of directories we're processing. This means there's also no upper bound on memory use, directory file descriptors held, etc. etc. etc.

Maybe we should take a step back and see how we can add parallelism without unbounded resource usage. What if you were to walk the directory hierarchy synchronously (like we already do right now), but at least add parallelism downloading/hardlinking files? Maybe that already brings performance close to what you desire without complicating the design too much.

EdSchouten:
Also, there is one refactoring that's still planned somewhere in the back of my queue: We should get rid of ContentAddressableStorage.GetFile()/PutFile(). Those are just silly. All of the other functions are for getting/putting stuff that can be located anywhere, while the *File() variants always stuff data in directories.

Maybe the code would be more elegant if we added a dedicated FileDownloader or something. Then you could have a BatchedFileDownloader, ParallelFileDownload, etc. etc. etc. without needing to care about the rest of ContentAddressableStorage.

pkg/builder/naive_build_directory.go
allada: aaagh, I'm not used to writing go, and I'm still thinking C++ lifetimes apply.

pkg/builder/naive_build_directory.go
allada: Hmmm, I'm not thrilled about the idea of restricting walking directories to be synchronous. I suppose it wouldn't be too bad to do them on a per-level basis, but I think this would still have the open folder problem since one folder could have hundreds of folders inside it, so no matter how you look at it, we will always have at least MAX_SLOTS folder descriptors open. There's also the problem of complexity this would induce. This PR is already adding a bit of complexity as it is which I'm not happy with at all.

Maybe we could do this in passes and instead of trying to place the files/folders immediately we place the files into a flat folder with hashes/indexes then once all of them are available we could materialize the folder tree by making each folder and moving each file into the retrospective directory?

There's also a few things we could do to significantly reduce the chance of hitting our limit (but not removing entirely), like only allowing one thread to process the "file" for-loop at a time. This way statistically we'd have most of our files clumped together along with a refcounter to know when it is safe to close the file descriptor of the parent folder.

I personally prefer the first option, but will mean we'd need to either abuse the `cache` folder structure or create another folder for this temp stuff (which I'd strongly object towards).

EdSchouten: > [...] we will always have at least MAX_SLOTS folder descriptors open. [...]

Wait. There's a subtle difference here.

With the solution that I'm proposing, the number of file descriptors that are open will be `O(k*d + n)`, where `k` is the concurrency of the worker, `d` is the maximum depth of an input root and `n` is the maximum slots for parallel uploading.

With the change that you're proposing, there is absolutely no bound at all. It's `O(∞)` (if that's even a thing).

Sure, there is a quadratic component in what I'm proposing and it's true what you state: it's at least MAX_SLOTS. Still, that's better than having absolutely no bound whatsoever.

allada: I don't think so, the complexity should be only `O(max(MAX_SLOTS, FOLDER_DEPTH))`.

It would follow something like:
* Create a temp folder
* Download each file into this directory (no sub-folders) with the key being the file's hash.
* After all files are downloaded walk the digest tree again and build the directory tree and move/hardlink the files to the new directory to the proper location.
* Delete temp folder

[we could use the same folder instead of creating a temp folder, but for illustration reasons I kept them separate].

FYI, we are not having issues with uploading from the worker [yet] but are having issues downloading to the worker. In our case it is because we have a folder that tens of thousands of actions use to compile (python) and this digest/folder has ~10k files inside it. Once a worker downloads it it is cached, but since we are using spot instances our workers die frequently, so we are trying to optimize this specific use case.

EdSchouten:
Closing this due to inactivity. Feel free to reopen if needed. Thanks!

PR #52, opened 2020-04-24
Add ability to install devices to input root

sdclarke:
This allows the worker to create device nodes in the dev directory of the input root of the action it is running. This is useful to have device nodes when chrooting.

Depends on buildbarn/bb-storage#58

### DIFF
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -173,6 +173,19 @@ func main() {
 				log.Fatal("Failed to parse maximum execution timeout")
 			}
 
+			var devices []filesystem.Device
+			for _, device := range runnerConfiguration.Devices {
+				var stat unix.Stat_t
+				if err := unix.Stat(filepath.Join("/dev", device), &stat); err != nil {
+					log.Fatalf("Unable to stat device /dev/%#v", device)
+				}
+				devices = append(devices, filesystem.Device{
+					Name: device,
+					Rdev: stat.Rdev,
+					Mode: os.FileMode(stat.Mode),
+				})
+			}
+
 			// Execute commands using a separate runner process. Due to the
 			// interaction between threads, forking and execve() returning
 			// ETXTBSY, concurrent execution of build actions can only be
@@ -241,19 +254,6 @@ func main() {
 					workerName, err := json.Marshal(workerID)
 					if err != nil {
 						log.Fatal("Failed to marshal worker ID: ", err)
-					}
-
-					var devices []filesystem.Device
-					for _, device := range configuration.Devices {
-						var stat unix.Stat_t
-						if err := unix.Stat(filepath.Join("/dev", device), &stat); err != nil {
-							log.Fatalf("Unable to stat device /dev/%#v", device)
-						}
-						devices = append(devices, filesystem.Device{
-							Name: device,
-							Rdev: stat.Rdev,
-							Mode: os.FileMode(stat.Mode),
-						})
 					}
 
 					buildExecutor := builder.NewLoggingBuildExecutor(
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strconv"
 	"syscall"
 	"time"
@@ -34,6 +35,8 @@ import (
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/gorilla/mux"
+
+	"golang.org/x/sys/unix"
 )
 
 func main() {
@@ -240,6 +243,19 @@ func main() {
 						log.Fatal("Failed to marshal worker ID: ", err)
 					}
 
+					var devices []filesystem.Device
+					for _, device := range configuration.Devices {
+						var stat unix.Stat_t
+						if err := unix.Stat(filepath.Join("/dev", device), &stat); err != nil {
+							log.Fatalf("Unable to stat device /dev/%#v", device)
+						}
+						devices = append(devices, filesystem.Device{
+							Name: device,
+							Rdev: stat.Rdev,
+							Mode: os.FileMode(stat.Mode),
+						})
+					}
+
 					buildExecutor := builder.NewLoggingBuildExecutor(
 						builder.NewCachingBuildExecutor(
 							builder.NewMetricsBuildExecutor(
@@ -252,7 +268,8 @@ func main() {
 												runner.NewRemoteRunner(runnerConnection),
 												clock.SystemClock,
 												defaultExecutionTimeout,
-												maximumExecutionTimeout),
+												maximumExecutionTimeout,
+												devices),
 											contentAddressableStorageFlusher),
 										clock.SystemClock,
 										string(workerName)))),
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -173,17 +173,17 @@ func main() {
 				log.Fatal("Failed to parse maximum execution timeout")
 			}
 
-			var devices []filesystem.Device
-			for _, device := range runnerConfiguration.Devices {
+			inputRootCharacterDevices := make(map[string]int)
+			for _, device := range runnerConfiguration.InputRootCharacterDeviceNodes {
 				var stat unix.Stat_t
-				if err := unix.Stat(filepath.Join("/dev", device), &stat); err != nil {
-					log.Fatalf("Unable to stat device /dev/%#v", device)
+				devicePath := filepath.Join("/dev", device)
+				if err := unix.Stat(devicePath, &stat); err != nil {
+					log.Fatalf("Unable to stat character device %#v: %s", devicePath, err)
 				}
-				devices = append(devices, filesystem.Device{
-					Name: device,
-					Rdev: stat.Rdev,
-					Mode: os.FileMode(stat.Mode),
-				})
+				if stat.Mode&syscall.S_IFMT != syscall.S_IFCHR {
+					log.Fatalf("The specified device %#v is not a character device", devicePath)
+				}
+				inputRootCharacterDevices[device] = int(stat.Rdev)
 			}
 
 			// Execute commands using a separate runner process. Due to the
@@ -269,7 +269,7 @@ func main() {
 												clock.SystemClock,
 												defaultExecutionTimeout,
 												maximumExecutionTimeout,
-												devices),
+												inputRootCharacterDevices),
 											contentAddressableStorageFlusher),
 										clock.SystemClock,
 										string(workerName)))),
--- a/pkg/builder/local_build_executor.go
+++ b/pkg/builder/local_build_executor.go
@@ -59,6 +59,10 @@ func (be *localBuildExecutor) uploadDirectory(ctx context.Context, outputDirecto
 	for _, file := range files {
 		name := file.Name()
 		childComponents := append(components, name)
+		// The elided default case of the below switch statement would
+		// represent a UNIX socket, FIFO or device node. These files cannot
+		// be represented in a Directory message. They are simply skipped.
+		// Returning an error would make the overall user experience worse.
 		switch fileType := file.Type(); fileType {
 		case filesystem.FileTypeRegularFile, filesystem.FileTypeExecutableFile:
 			childDigest, err := be.contentAddressableStorage.PutFile(ctx, outputDirectory, name, parentDigest)
@@ -106,10 +110,6 @@ func (be *localBuildExecutor) uploadDirectory(ctx context.Context, outputDirecto
 				Name:   name,
 				Target: target,
 			})
-		case filesystem.FileTypeOther:
-			continue
-		default:
-			return nil, status.Errorf(codes.Internal, "Output file %#v is not a regular file, directory or symlink", name)
 		}
 	}
 	return &directory, nil
@@ -156,18 +156,18 @@ func (be *localBuildExecutor) createOutputParentDirectory(inputRootDirectory fil
 	return d, nil
 }
 
-func (be *localBuildExecutor) createDev(inputRootDirectory BuildDirectory) error {
+func (be *localBuildExecutor) createCharacterDevices(inputRootDirectory BuildDirectory) error {
 	if err := inputRootDirectory.Mkdir("dev", 0777); err != nil && !os.IsExist(err) {
-		return util.StatusWrap(err, "Unable to create dev directory in input root")
+		return util.StatusWrap(err, "Unable to create /dev directory in input root")
 	}
 	devDir, err := inputRootDirectory.EnterDirectory("dev")
 	defer devDir.Close()
 	if err != nil {
-		return util.StatusWrap(err, "Unable to enter dev directory in input root")
+		return util.StatusWrap(err, "Unable to enter /dev directory in input root")
 	}
 	for name, number := range be.inputRootCharacterDevices {
-		if err := devDir.Mknod(name, os.FileMode(os.ModeDevice|os.ModeCharDevice|0666), number); err != nil {
-			return util.StatusWrapf(err, "Mknod failed for device %#v", name)
+		if err := devDir.Mknod(name, os.ModeDevice|os.ModeCharDevice|0666, number); err != nil {
+			return util.StatusWrapf(err, "Failed to create character device %#v", name)
 		}
 	}
 	return nil
@@ -270,7 +270,7 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 	}
 
 	if len(be.inputRootCharacterDevices) > 0 {
-		if err := be.createDev(inputRootDirectory); err != nil {
+		if err := be.createCharacterDevices(inputRootDirectory); err != nil {
 			attachErrorToExecuteResponse(response, err)
 			return response
 		}
--- a/pkg/builder/local_build_executor.go
+++ b/pkg/builder/local_build_executor.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"path"
 	"strings"
-	"syscall"
 	"time"
 
 	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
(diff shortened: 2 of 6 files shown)


cmd/bb_worker/main.go
```diff
@@ -232,6 +235,19 @@ func main() {
...
+				var devices []filesystem.Device
+				for _, device := range configuration.Devices {
+					var stat unix.Stat_t
+					if err := unix.Stat(filepath.Join("/dev", device), &stat); err != nil {
+						log.Fatalf("Unable to stat device /dev/%#v", device)
+					}
+					devices = append(devices, filesystem.Device{
+						Name: device,
+						Rdev: stat.Rdev,
+						Mode: os.FileMode(stat.Mode),
```
EdSchouten, line 247: Inheriting whether something is a block or character device and its rdev number make sense. I do wonder whether permissions should be inherited. That’s unrelated, right?

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -50,6 +50,13 @@ message ApplicationConfiguration {
 
   // Common configuration options that apply to all Buildbarn binaries.
   buildbarn.configuration.global.Configuration global = 19;
+
+  // Install devices into /dev of the input root. Useful in combination
+  // with having the runner chroot into the input root.
+  bool install_devices = 20;
```
EdSchouten, line 56: Do we need the additional boolean flag? Doesn’t a non-empty devices list already imply the boolean value?

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -50,6 +50,13 @@ message ApplicationConfiguration {
...
+
+  // Device nodes to install in input root
```
EdSchouten, line 58: (Nit: End all sentences with a full stop.)

This should likely contain some example, right? Otherwise it becomes ambiguous whether the /dev/ prefix needs to be included.

Also, this is currently configured at the global level, even though I think it would make far more sense to be able to specify this at the runner level. I can well imagine a multi-purpose worker that does chrooting for a certain runner, while doing plain Ubuntu container builds on another runner.

sdclarke:
I need to add some logic to avoid trying to upload the device nodes if they would be included in an output directory requested by the action. For the sake of testing I added a `continue` if the file type was `FileTypeOther` [here](https://github.com/buildbarn/bb-remote-execution/blob/14971580d7ad230909118a53f582aa1f16ff001e/pkg/builder/local_build_executor.go#L106), but this is obviously not a viable solution for the long term.

pkg/builder/local_build_executor.go
```diff
@@ -104,6 +107,8 @@ func (be *localBuildExecutor) uploadDirectory(ctx context.Context, outputDirecto
 				Name:   name,
 				Target: target,
 			})
+		case filesystem.FileTypeCharacterDevice:
```
EdSchouten, line 110: Oh, wait. You do check for the file type here.

Question: What makes character devices so special in this case? What about block devices? Maybe it would be acceptable to simply skip any non-file/device/symlink? I can imagine that FIFOs, UNIX sockets, etc. inside of output directories are also problematic.

pkg/builder/local_build_executor.go
```diff
@@ -32,18 +32,21 @@ type localBuildExecutor struct {
...
-func NewLocalBuildExecutor(contentAddressableStorage cas.ContentAddressableStorage, buildDirectoryCreator BuildDirectoryCreator, runner runner.Runner, clock clock.Clock, defaultExecutionTimeout time.Duration, maximumExecutionTimeout time.Duration) BuildExecutor {
+func NewLocalBuildExecutor(contentAddressableStorage cas.ContentAddressableStorage, buildDirectoryCreator BuildDirectoryCreator, runner runner.Runner, clock clock.Clock, defaultExecutionTimeout time.Duration, maximumExecutionTimeout time.Duration, devices []re_filesystem.Device) BuildExecutor {
 	return &localBuildExecutor{
 		contentAddressableStorage: contentAddressableStorage,
 		buildDirectoryCreator:     buildDirectoryCreator,
 		runner:                    runner,
 		clock:                     clock,
 		defaultExecutionTimeout:   defaultExecutionTimeout,
-		maximumExecutionTimeout:   maximumExecutionTimeout,
+
```
EdSchouten, line 47: Nit: Unnecessary empty line.

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -113,4 +113,8 @@ message RunnerConfiguration {
   // as announced to the scheduler. In addition to the labels specified,
   // one named "thread" is added automatically in case concurrency > 1.
   map<string, string> worker_id = 8;
+
+  // Device nodes to install in input root.
+  // e.g. ["null", "zero", "random"]
+  repeated string devices = 9;
```
EdSchouten, line 119: Nit: Maybe call this `input_root_device_nodes`? `devices` might be a bit too generic.

cmd/bb_worker/main.go
```diff
@@ -170,6 +173,18 @@ func main() {
 				log.Fatal("Failed to parse maximum execution timeout")
 			}
 
+			var devices []re_filesystem.Device
+			for _, device := range runnerConfiguration.Devices {
+				var stat unix.Stat_t
+				if err := unix.Stat(filepath.Join("/dev", device), &stat); err != nil {
+					log.Fatalf("Unable to stat device /dev/%#v", device)
```
EdSchouten, line 180: I suspect that this will print `Unable to stat device /dev/"null"`. It also doesn't print `err`. What are your thoughts on doing something like this?
```go
devicePath := filepath.Join("/dev", device)
if err := unix.Stat(devicePath), &stat); err != nil {
        log.Fatalf("Unable to stat device %#v: %s", devicePath, err)
}
```

pkg/builder/local_build_executor.go
```diff
@@ -32,18 +32,21 @@ type localBuildExecutor struct {
 	clock                     clock.Clock
 	defaultExecutionTimeout   time.Duration
 	maximumExecutionTimeout   time.Duration
+	devices                   []re_filesystem.Device
 }
 
 // NewLocalBuildExecutor returns a BuildExecutor that executes build
 // steps on the local system.
-func NewLocalBuildExecutor(contentAddressableStorage cas.ContentAddressableStorage, buildDirectoryCreator BuildDirectoryCreator, runner runner.Runner, clock clock.Clock, defaultExecutionTimeout time.Duration, maximumExecutionTimeout time.Duration) BuildExecutor {
+func NewLocalBuildExecutor(contentAddressableStorage cas.ContentAddressableStorage, buildDirectoryCreator BuildDirectoryCreator, runner runner.Runner, clock clock.Clock, defaultExecutionTimeout time.Duration, maximumExecutionTimeout time.Duration, devices []re_filesystem.Device) BuildExecutor {
```
EdSchouten, line 40: Maybe it would make more sense to let this be a `map[string]int`?

pkg/builder/local_build_executor.go
```diff
@@ -248,6 +253,13 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 		return response
 	}
 
+	if be.devices != nil {
```
EdSchouten, line 256: I would suggest that `CreateDev()` is simply inlined inside this function. Even though it mutates the file system, it's not generic enough that it warrants having it as a utility function in `re_filesystem`.

pkg/builder/local_build_executor_test.go
```diff
@@ -33,7 +33,7 @@ func TestLocalBuildExecutorInvalidActionDigest(t *testing.T) {
 	buildDirectoryCreator := mock.NewMockBuildDirectoryCreator(ctrl)
```
EdSchouten, line 33: Could you add some unit testing coverage for device node creation as well?

pkg/filesystem/device_darwin.go
```diff
@@ -0,0 +1,23 @@
...
+
+// Device stores the name, device number and mode of a device to be used
+// when creating device nodes in the input root dev directory
+type Device struct {
+	Name string
+	Rdev int32
+}
+
+// CreateDev will return a fixed error on darwin
+func CreateDev(inputRootDirectory filesystem.Directory, devices []Device) error {
```
EdSchouten, line 21: I don't think we really need this stub on macOS. Directory.Mknod() will already fail on that platform.

pkg/filesystem/device_linux.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	devDir, err := inputRootDirectory.EnterDirectory("dev")
+	if err != nil {
+		return util.StatusWrap(err, "Unable to enter dev directory in input root")
+	}
+	for _, device := range devices {
+		if _, err := devDir.Lstat(device.Name); err == nil {
+			continue
+		}
+		if err := devDir.Mknod(device.Name, os.FileMode(syscall.S_IFCHR|0666), int(device.Rdev)); err != nil {
```
EdSchouten, line 34: So this is just about getting character devices working, right? Not block devices. Do we care about block devices?

If we don't care about block devices, could we change the phrasing everywhere to use 'character device' instead of just 'device'? If we do care about block devices, then we should let main.go save the file type as well.

Note that I'm fine with just supporting character devices. If we limit ourselves to that, main.go should have a check to enforce that as well.

pkg/builder/local_build_executor.go
sdclarke: Nothing really makes them special I suppose, they were just the only scope of this PR and I didn't want to just remove the failure state for `FileTypeOther` and replace it with a `continue`. If that turns out to be the more desirable solution, or if there's something else that should be done instead I'm more than happy to change it up.

pkg/filesystem/device_linux.go
sdclarke: I believe we do only care about character devices in this case, so I'll go about making that more explicit throughout.

pkg/builder/local_build_executor.go
sdclarke: That would presumably allow us to dispense with the `re_filesystem.Device` struct?

pkg/builder/local_build_executor_test.go
sdclarke: Do I need to somehow let the test create actual device nodes, or is it enough to simply mock the directory `<input root>/dev` and check the `Mknod` calls are made against it when there is a non-empty map of device nodes passed to `NewLocalBuildExecutor()`?

sdclarke: And if it's the second case can I do something like in #56 where I just modified an existing test?

EdSchouten: Testing the creation of actual device nodes is likely impossible, because that would require the unit tests to run with super-user rights, which is impractical. Mocking the "dev" directory and capturing Mknod() calls is all right>

With regards to your second question: for the success case it's sufficient to just modify one of the existing tests. You likely want to test the failure case as well (when Mknod() fails). This should better go in its own test.

pkg/builder/local_build_executor.go
EdSchouten: Exactly!

pkg/builder/local_build_executor.go
EdSchouten: Skipping all `FileTypeOther` entries is fine in my opinion.

pkg/builder/local_build_executor.go
```diff
@@ -248,6 +253,24 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 		return response
 	}
 
+	if len(be.inputRootCharacterDevices) > 0 {
+		if err := inputRootDirectory.Mkdir("dev", 0777); err != nil && !os.IsExist(err) {
+			attachErrorToExecuteResponse(response, util.StatusWrap(err, "Unable to create dev directory in input root"))
```
EdSchouten, line 258: Any error prior to execution should lead to direct termination. Don't forget to `return response` here!

pkg/builder/local_build_executor.go
```diff
@@ -248,6 +253,24 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
...
+		}
+		devDir, err := inputRootDirectory.EnterDirectory("dev")
```
EdSchouten, line 260: This code currently leaks `devDir`. If it makes your life easier, just add a helper function to `localBuildExecutor`, so you can use `defer devDir.Close()` there to close the directory afterwards.

pkg/builder/local_build_executor.go
```diff
@@ -248,6 +253,24 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
...
+		if err != nil {
+			attachErrorToExecuteResponse(response, util.StatusWrap(err, "Unable to enter dev directory in input root"))
+		}
+		for name, number := range be.inputRootCharacterDevices {
+			if _, err := devDir.Lstat(name); err == nil {
```
EdSchouten, line 265: Is this check needed?

pkg/builder/local_build_executor.go
```diff
@@ -248,6 +253,24 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
...
+				continue
+			}
+			if err := devDir.Mknod(name, os.FileMode(syscall.S_IFCHR|0666), number); err != nil {
```
EdSchouten, line 268: This should likely use `os.ModeDevice | os.ModeCharDevice | 0666`.

pkg/builder/local_build_executor.go
sdclarke: I think if that isn't there and the file already exists the Mknod call will return an error.

EdSchouten: But that's not necessarily a bad thing, right?

pkg/builder/local_build_executor.go
sdclarke: That creates a different value which I'm fairly sure doesn't create a device node when used with Mknod as I tried it before:
```
2020/05/28 10:24:16 mode|os.ModeDevice|os.ModeCharDevice: 0x42001b6
2020/05/28 10:24:16 mode|syscall.S_IFCHR: 0x21b6
```

EdSchouten: Sure, but what I’m getting at is that S_IFCHR is not a value that you’re normally supposed to store in an os.FileMode. We either use os.FileMode in combination with the intended numerical constants, or we don’t.

sdclarke: So it may be best to not have Mknod take `os.FileMode` as and argument and just let it be `uint32` directly? Then I could just not cast to `os.FileMode` and pass `syscall.S_IFCHR|0666` directly?

EdSchouten: Though that is one option to do it, I would suggest we stick to using os.FileMode and let this code pass in the os.Mode* constants. That avoids a dependency on sys/unix in this code. Having such a dependency may be problematic if we ever want to make Windows support more complete.

sdclarke: Or pass the `os.Mode...` values as you suggested and handle it on the other end

pkg/builder/local_build_executor.go
```diff
@@ -104,6 +106,8 @@ func (be *localBuildExecutor) uploadDirectory(ctx context.Context, outputDirecto
 				Name:   name,
 				Target: target,
 			})
+		case filesystem.FileTypeOther:
```
EdSchouten, line 109: It's annoying that Go doesn't have true native enum types and that `switch` can be tagged to require that all cases are handled.

Aaanyway: I propose that we replace all of this by just:
```go
default:
    // A UNIX socket, FIFO or device node. These files cannot be represented in
    // a Directory message. Simply skip them. Returning an error here would
    // make the overall user experience worse.
    continue
```

pkg/builder/local_build_executor.go
```diff
@@ -152,6 +156,23 @@ func (be *localBuildExecutor) createOutputParentDirectory(inputRootDirectory fil
 	return d, nil
 }
 
+func (be *localBuildExecutor) createDev(inputRootDirectory BuildDirectory) error {
```
EdSchouten, line 159: `createCharacterDevices()`?

pkg/builder/local_build_executor.go
```diff
@@ -152,6 +156,23 @@ func (be *localBuildExecutor) createOutputParentDirectory(inputRootDirectory fil
...
+	if err := inputRootDirectory.Mkdir("dev", 0777); err != nil && !os.IsExist(err) {
+		return util.StatusWrap(err, "Unable to create dev directory in input root")
```
EdSchouten, line 161: Even though the thing is called `dev`, maybe it's good to use `/dev` in error messages? That makes the intent a bit more clear.

pkg/builder/local_build_executor.go
```diff
@@ -152,6 +156,23 @@ func (be *localBuildExecutor) createOutputParentDirectory(inputRootDirectory fil
...
+	}
+	devDir, err := inputRootDirectory.EnterDirectory("dev")
+	defer devDir.Close()
+	if err != nil {
+		return util.StatusWrap(err, "Unable to enter dev directory in input root")
+	}
+	for name, number := range be.inputRootCharacterDevices {
+		if err := devDir.Mknod(name, os.FileMode(os.ModeDevice|os.ModeCharDevice|0666), number); err != nil {
```
EdSchouten, line 169: Do you still need the `os.FileMode()` here?

pkg/builder/local_build_executor.go
```diff
@@ -152,6 +156,23 @@ func (be *localBuildExecutor) createOutputParentDirectory(inputRootDirectory fil
...
+			return util.StatusWrapf(err, "Mknod failed for device %#v", name)
```
EdSchouten, line 170: `"Failed to create character device %#v"`? 'Mknod' is a bit too technical here in my opinion.

pkg/builder/local_build_executor.go
EdSchouten: Or wait. We don't even need the `default: continue`. There is nothing after this switch statement, so there's no need to `continue`. In that case: just place a similar comment above the switch statement.

PR #54, opened 2020-05-10
Create bb-runner-installer image

asuffield:
This image is for use as an init container on kubernetes, to allow
bb-runner to be used as a mixin to existing containers. Notably, this
allows users to pick any RBE image version, not just the one which
bb-remote-execution is currently selecting.

This turns out to be important bacause bazel-toolchains selects image
versions based on the version of bazel being run, and there is no
image which is compatible with all bazel versions.

### DIFF
--- a/cmd/bb_runner/BUILD.bazel
+++ b/cmd/bb_runner/BUILD.bazel
@@ -54,8 +54,8 @@ container_layer(
 container_image(
     name = "bb_runner_installer",
     base = "@busybox//image",
-    entrypoint = ["/install"],
     cmd = ["/bb"],
+    entrypoint = ["/install"],
     layers = [
         ":install_layer",
         ":bb_runner_layer",
--- a/cmd/bb_runner/BUILD.bazel
+++ b/cmd/bb_runner/BUILD.bazel
@@ -46,6 +46,24 @@ container_layer(
     files = ["@com_github_krallin_tini_tini_static_amd64//file"],
 )
 
+container_layer(
+    name = "install_layer",
+    files = ["install"],
+)
+
+container_image(
+    name = "bb_runner_installer",
+    base = "@busybox//image",
+    entrypoint = ["/install"],
+    cmd = ["/bb"],
+    layers = [
+        ":install_layer",
+        ":bb_runner_layer",
+        ":tini_layer",
+    ],
+    visibility = ["//visibility:public"],
+)
+
 container_image(
     name = "bb_runner_ubuntu16_04_container",
     base = "@rbe_ubuntu16_04_base//image",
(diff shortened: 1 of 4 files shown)


cmd/bb_runner/BUILD.bazel
```diff
@@ -46,6 +46,24 @@ container_layer(
...
+
+container_image(
+    name = "bb_runner_installer",
+    base = "@busybox//image",
+    cmd = ["/bb"],
+    entrypoint = ["/install"],
+    layers = [
+        ":install_layer",
+        ":bb_runner_layer",
+        ":tini_layer",
```
EdSchouten, line 62: Oh man. Life would eventually be so much nicer if we didn't need tini. Right now it's only there to prevent zombie process leakage. Reaping those zombies could trivially be done by bb_runner, if it weren't for the fact that that wouldn't be able to interact with cmd.Run() properly. :-/

cmd/bb_runner/install
```diff
@@ -0,0 +1,11 @@
...
+
+set -eu
+
+echo "$*"
+
+# Can pass the destination directory as the first argument. By default
+# it will use /bb
+dest="${1:-/bb}"
+
+cp /bb_runner /tini "$dest"
```
EdSchouten, line 11: Maybe we should place all these files in a subdirectory, so they can be copied into into /bb without requiring constant changes to this script?

cmd/bb_runner/BUILD.bazel
```diff
@@ -46,6 +46,24 @@ container_layer(
...
+    files = ["install"],
+)
+
+container_image(
+    name = "bb_runner_installer",
+    base = "@busybox//image",
+    cmd = ["/bb"],
+    entrypoint = ["/install"],
+    layers = [
+        ":install_layer",
```
EdSchouten, line 60: What about the passwd layer? Wouldn't we want to bring that one in as well? Or do we expect people to run builds as some other existing user (e.g., `nobody`)?

asuffield: On kubernetes it doesn't seem to be important, and splicing files into /etc (which exists in all the underlying images) is much trickier, we can't just mount over it and it doesn't exist yet when the init container runs. The RBE images have a bunch of users in them already, and other images may have their own user field in the docker manifest - since we're not supplying the container image itself in this model, we can't set it directly. We can only directly influence the user in the init container, which isn't interesting.

Some brief inspection indicates that the ubuntu RBE images come up as uid=0 by default, but we could pick nobody or bin or something in the k8s config.

My considered opinion on this topic is that only the thing which builds the underlying runner image has any hope of solving this problem, so we just stop trying to solve it. In this model bb_runner is a guest injected into this operating system.

cmd/bb_runner/install
asuffield: I thought about this, and then I thought "how often do we really add new binaries to bb_runner?"

Notably it would require changing the bb_runner and tini layers, so I'd have to duplicate those to avoid changing how the ubuntu image works, which seemed like more trouble than it's worth for a thing that might never happen.

PR #55, opened 2020-05-18
Allow setting of file mtimes

sdclarke:
Add option to set mtimes of all files to 2011-11-11 11:11:11 when they are hardlinked into the build directory.

Requires buildbarn/bb-storage#60 in order to work.

### DIFF
--- a/pkg/cas/hardlinking_content_addressable_storage.go
+++ b/pkg/cas/hardlinking_content_addressable_storage.go
@@ -3,20 +3,24 @@ package cas
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/buildbarn/bb-storage/pkg/cas"
 	"github.com/buildbarn/bb-storage/pkg/digest"
 	"github.com/buildbarn/bb-storage/pkg/eviction"
 	"github.com/buildbarn/bb-storage/pkg/filesystem"
+
+	"golang.org/x/sys/unix"
 )
 
 type hardlinkingContentAddressableStorage struct {
 	cas.ContentAddressableStorage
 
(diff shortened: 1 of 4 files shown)


pkg/cas/hardlinking_content_addressable_storage.go
```diff
@@ -105,6 +110,15 @@ func (cas *hardlinkingContentAddressableStorage) GetFile(ctx context.Context, di
 		if err := directory.Link(name, cas.cacheDirectory, key); err != nil {
 			return err
 		}
+		if cas.deterministicMtimes {
```
EdSchouten, line 113: Instead of fixing this up in the hardlinking cache, would it make more sense to do this as part of bb-storage/pkg/cas/blob_access_content_addressable_storage.go? That's where the files get created in the first place.

Because that function also happens to have a reference to the file descriptor that's used to create the file, we could even decide to use `futimens()` instead of `utimensat()` there.

EdSchouten: To clarify, you could extend `FileAppender` to gain this function:
```
type FileAppender interface {
        ...
        Chtimes(atime time.Time, mtime time.Time) error
}
```

Then blobAccessContentAddressableStorage.GetFile() could call w.Chtimes() after creating the file has succeeded.

I think that approach is simple and elegant enough that it doesn't necessarily need its own configuration option. Feel free to just enable that unconditionally.

sdclarke:
Given the discussion in this pull request it is being closed as its functionality will be incorporated into buildbarn/bb-storage#60.

PR #56, opened 2020-05-27
Allow a nil platform message

sdclarke:
This makes the scheduler allow commands to not specify a platform
message, and it will interpret the lack of platform message in the same
way as an empty list of key/value pairs

### DIFF
--- a/pkg/builder/in_memory_build_queue_test.go
+++ b/pkg/builder/in_memory_build_queue_test.go
@@ -158,34 +158,6 @@ func TestInMemoryBuildQueueExecuteBadRequest(t *testing.T) {
 		require.Equal(t, err, status.Error(codes.FailedPrecondition, "Failed to obtain command: Blob not found"))
 	})
 
-	// Command does not contain any playform requirements.
-	t.Run("MissingPlatform", func(t *testing.T) {
-		contentAddressableStorage.EXPECT().GetAction(
-			gomock.Any(),
-			digest.MustNewDigest("main", "da39a3ee5e6b4b0d3255bfef95601890afd80709", 123),
-		).Return(&remoteexecution.Action{
-			CommandDigest: &remoteexecution.Digest{
-				Hash:      "61c585c297d00409bd477b6b80759c94ec545ab4",
-				SizeBytes: 456,
-			},
-		}, nil)
-		contentAddressableStorage.EXPECT().GetCommand(
-			gomock.Any(),
-			digest.MustNewDigest("main", "61c585c297d00409bd477b6b80759c94ec545ab4", 456),
-		).Return(&remoteexecution.Command{}, nil)
-
-		stream, err := executionClient.Execute(ctx, &remoteexecution.ExecuteRequest{
-			InstanceName: "main",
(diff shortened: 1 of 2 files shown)


pkg/builder/in_memory_build_queue_test.go
```diff
@@ -158,34 +158,6 @@ func TestInMemoryBuildQueueExecuteBadRequest(t *testing.T) {
 		require.Equal(t, err, status.Error(codes.FailedPrecondition, "Failed to obtain command: Blob not found"))
 	})
 
-	// Command does not contain any playform requirements.
```
EdSchouten, line 161: Ha, typo in there!

Instead of removing this test, could we either change it (or add a new test) to validate the new behaviour?

sdclarke: I guess it would be best to move out from under the `BadRequest` test? Or maybe it still fits there as a missing platform message could still be considered a bad request?

EdSchouten: Most of the tests for the correct cases below use the same platform properties. I think it's acceptable if you just took one of those tests and modified it to use a nil platform message.

sdclarke: Do you mean just take any one of the existing tests, such as `TestInMemoryBuildQueuePurgeStaleWorkersAndQueues`, and change the platform message to be nil? Or to create a new test which beaves similarly to one of the other tests but with a nil platform?

I tried leaving it in the same place and checking that the error is that no workers exist rather than the platform isn't specified, but that is probably not a great idea as it's not isolating one the part of the code being tested.

EdSchouten: The first. Just adjust one of the existing ones. :-)

pkg/builder/in_memory_build_queue_test.go
```diff
@@ -558,7 +504,7 @@ func TestInMemoryBuildQueuePurgeStaleWorkersAndQueues(t *testing.T) {
 	})
 	require.NoError(t, err)
 	_, err = stream3.Recv()
-	require.Equal(t, err, status.Error(codes.FailedPrecondition, "No workers exist for instance \"main\" platform {\"properties\":[{\"name\":\"cpu\",\"value\":\"armv6\"},{\"name\":\"os\",\"value\":\"linux\"}]}"))
+	require.Equal(t, err, status.Error(codes.FailedPrecondition, "No workers exist for instance \"main\" platform "))
```
EdSchouten, line 507: Ah, interesting. We just learned that a nil message marshals to the empty string; not `{}`. That sounds like something that's going to confuse people.

Maybe it would be better to do what you proposed on Slack? Revert the changes to in_memory_build_queue.go. Then change the error path to `platform = &remoteexecution.Platform{}`. That way the resulting string is guaranteed to be non-empty.

EdSchouten: Now that I think of it, I guess we must also do this to make the scheduler web UI work properly. It becomes hard to click on the platform queue name if it's an empty string.

PR #60, opened 2020-06-11
Add a note about the shared blobstore on frontend and workers

vors:
A follow-up from the conversation on buildteam slack.

I'm not sure this is the best place for this documentation, please advice if there is a better one.
I think it's discoverable here, at least for this particular scenario (`forward_metadata`).

### DIFF
--- a/pkg/proto/configuration/bb_worker/bb_worker.proto
+++ b/pkg/proto/configuration/bb_worker/bb_worker.proto
@@ -13,6 +13,15 @@ option go_package = "github.com/buildbarn/bb-remote-execution/pkg/proto/configur
 
 message ApplicationConfiguration {
   // Configuration for blob storage.
+  //
+  // Note: some configurations have implmentation nounces that lead to a differences in runtime behavior
+  // between bb-storage acting as a bb-frontend and bb-worker, even while blobstore configuration is shared.
+  ///
+  // For example: buildbarn.configuration.grpc.ClientConfiguration.forward_metadata = ["authorization"]
+  // will work on the bb-storage frontend and forward client provided "authorization" header.
+  // In on the worker, the grpc call is terminated in the scheduler and when worker synchronizes with scheduler
+  // the client headers are not passed as part of this context.
+  // Hence the `forward_metadata` on the worker would not be able to re-use the client "authorization" header.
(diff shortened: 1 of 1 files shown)


pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -13,6 +13,15 @@ option go_package = "github.com/buildbarn/bb-remote-execution/pkg/proto/configur
 
 message ApplicationConfiguration {
   // Configuration for blob storage.
+  //
+  // Note: some configurations have implmentation nounces that lead to a differences in runtime behavior
+  // between bb-storage acting as a bb-frontend and bb-worker, even while blobstore configuration is shared.
+  ///
+  // For example: buildbarn.configuration.grpc.ClientConfiguration.forward_metadata = ["authorization"]
+  // will work on the bb-storage frontend and forward client provided "authorization" header.
+  // In on the worker, the grpc call is terminated in the scheduler and when worker synchronizes with scheduler
```
vors, line 22: I'm not 100% sure does bb-frontend passes the context to bb-scheduler or it's terminated in bb-frontend.

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -13,6 +13,15 @@ option go_package = "github.com/buildbarn/bb-remote-execution/pkg/proto/configur
 
 message ApplicationConfiguration {
   // Configuration for blob storage.
+  //
+  // Note: some configurations have implmentation nounces that lead to a differences in runtime behavior
```
EdSchouten, line 17: I don't think it is right to document this here, as this is in no way specific to the blobstore API. This applies to any form of gRPC traffic. I also wouldn't describe it as an implementation nuance; it's inherent to how gRPC metadata forwarding works.

Wouldn't it make more sense to extend the description of `forward_metadata` to mention something like this?

> Keep in mind that this option can only be used to forward metadata from incoming calls on gRPC servers to outgoing calls on gRPC clients. This means that this option cannot be used to let bb-scheduler forward client provided credentials to workers.

Also note that the .proto files currently use a column limit of 80-ish characters.

vors: Makes sense, let me just copy verbatim your comment and send a PR to bb-storage.

PR #61, opened 2020-06-15
bb_worker: wait for bb_scheduler to be ready before initialising

mostynb:
This change makes bb_worker wait for bb_scheduler to be ready for up to 10s before it continues initialisation. This should avoid most "Failed to synchronize with scheduler" warnings at startup.

We check for readiniess by using the Capabilities service as a simple "ping" with no side effects.

### DIFF
--- a/cmd/bb_scheduler/main.go
+++ b/cmd/bb_scheduler/main.go
@@ -94,6 +94,9 @@ func main() {
 			bb_grpc.NewGRPCServersFromConfigurationAndServe(
 				configuration.WorkerGrpcServers,
 				func(s *grpc.Server) {
+					// So workers can check if we're ready:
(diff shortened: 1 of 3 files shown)


EdSchouten:
bb_worker and bb_scheduler are two separate services that shouldn't have any startup dependency on each other. Even if the scheduler is down, workers should go in a state that they have fully started. Also, the worker and scheduler can be restarted independently. Once the scheduler gets restarted later on, these error messages would still appear.

This change simply adds code that has no practical advantages. I am generally not in favour of those kinds of changes.

mostynb:
I think there's some value in not bothering admins with warnings that are expected but harmless (like when starting both services at about the same time). But no matter, it's not super important.

PR #66, opened 2020-08-02
Propagate trace context to workers

andyleap:
Basic proof of concept for propagating trace context from schedulers to workers.

### DIFF
--- a/pkg/builder/build_client.go
+++ b/pkg/builder/build_client.go
@@ -71,7 +71,6 @@ func (bc *BuildClient) startExecution(executionRequest *remoteworker.DesiredStat
 	// Spawn the execution of the build action.
 	var ctx context.Context
 	ctx, bc.executionCancellation = context.WithCancel(context.Background())
-
 	updates := make(chan *remoteworker.CurrentState_Executing, 10)
 	bc.executionUpdates = updates
 	go func() {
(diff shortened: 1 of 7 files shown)


pkg/builder/build_client.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/util"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
+	"go.opencensus.io/trace"
```
EdSchouten, line 16: Nit: imports in the Buildbarn tree are grouped by 2nd-level domain name, sorted alphabetically. These two should go below the golang.org ones.

pkg/builder/in_memory_build_queue.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
+	"go.opencensus.io/trace"
```
EdSchouten, line 28: Same here

pkg/builder/build_client.go
andyleap: That makes more sense, fixed!

PR #67, opened 2020-08-20
Add trace span details

asuffield:
This inserts trace spans for the waitExecution stream, and annotates
the Execution spans with the details of what happened.

Annotating with response.Message in particularly powerful here because
that happens to contain the bb-browser URL corresponding to the
execution, which is sufficient to explore the details of the
execution. The rest of the attributes are primarily useful for
searching for the right trace.

### DIFF
--- a/pkg/builder/build_client.go
+++ b/pkg/builder/build_client.go
@@ -75,20 +75,31 @@ func (bc *BuildClient) startExecution(executionRequest *remoteworker.DesiredStat
 	updates := make(chan *remoteworker.CurrentState_Executing, 10)
 	bc.executionUpdates = updates
 	go func() {
+		var span *trace.Span
 		if tc, ok := propagation.FromBinary(executionRequest.TraceContext); ok {
-			var span *trace.Span
 			ctx, span = trace.StartSpanWithRemoteParent(ctx, "BuildClient.Execute", tc)
 			defer span.End()
 		}
+
+		response := bc.buildExecutor.Execute(
+			ctx,
+			bc.filePool,
+			bc.instanceName,
+			executionRequest,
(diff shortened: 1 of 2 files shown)


EdSchouten:
I think this is related to what we discussed on Slack the other day. In my opinion it's a waste of effort to add all sorts of trace span details for things that are already embedded in Protobufs and sent across the wire using gRPC. There surely must be a generic way to expose information like this without explicitly annotating an entire codebase. Can't gRPC/Open{Tracing,Census,Telemetry} do this kind of stuff for us?

Why isn't there some gRPC middleware that we can insert that allows us to list which Protobuf fields we want to capture?

asuffield:
I'm sure it's possible to write the same code using an interceptor, or to write it in a language which isn't go which is executed by some interceptor. It will turn out to be exactly the same code though, so all we'd be doing is moving it around.

This PR does contain some subtleties which I can split out from the attributes if you like - the SetStatus on execution requests is mapping the somewhat weird error reporting of the remote execution API back onto the standard span fields, and the waitExecution change is creating spans that would not otherwise exist, because streaming messages don't produce spans.

asuffield:
I've filed https://github.com/open-telemetry/opentelemetry-go-contrib/issues/299 for the question of whether it's feasible to move this code into the generic middleware. I'm not optimistic for this being feasible, but we can explore it.

asuffield:
The interesting parts of this have migrated into https://github.com/buildbarn/bb-remote-execution/pull/69.

Unsurprisingly otel have kicked the question of trying to solve this generically into the long grass. I wouldn't expect anybody to solve that in the near future, it sounds hard.

PR #69, opened 2020-08-27
Convert to OpenTelemetry

asuffield:
To better fit the otel model, a separate producer/consumer span is
produced for the interaction between bb_scheduler and bb_worker, and
the waitExecution spans are linked to it. This will result in tracing
implementations representing the deduplication accurately.

This won't build until the corresponding PR to bb-storage is
merged. There doesn't seem to be any neat way to do this transition -
I do it locally with a replace directive in go.mod.


pkg/builder/in_memory_build_queue.go
```diff
@@ -343,6 +359,7 @@ func (bq *InMemoryBuildQueue) Execute(in *remoteexecution.ExecuteRequest, out re
 		pq.wakeupNextWorker()
 		pq.operationsQueuedTotal.Inc()
 	}
+
```
EdSchouten, line 362: Nit: unnecessary whitespace change.

pkg/builder/build_client.go
```diff
@@ -13,12 +13,12 @@ import (
 	"github.com/buildbarn/bb-storage/pkg/util"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
+	"go.opentelemetry.io/otel/api/trace"
```
EdSchouten, line 16: Nit: These imports should go below the golang.org ones.

pkg/builder/in_memory_build_queue.go
```diff
@@ -311,14 +313,28 @@ func (bq *InMemoryBuildQueue) Execute(in *remoteexecution.ExecuteRequest, out re
...
+		// https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md
+		// to use their span format.
+		//
+		// Note that ctx here is local to producing the injection, we
+		// don't otherwise use it.
+		ctx, span := bq.tracer.Start(ctx, "bb_worker send",
+			trace.WithSpanKind(trace.SpanKindProducer),
+			trace.WithAttributes(
+				label.String("messaging.system", "bb_scheduler"),
+				label.String("messaging.destination", instanceName.String()),
```
EdSchouten, line 326: Picking the instance name here seems a bit arbitrary. Wouldn't it make more sense to pick the pair of [instance name, platform]? That identifies queues right now.

This code is also largely repeated in the scheduler and the worker. Could it be factored out?

pkg/builder/in_memory_build_queue.go
```diff
@@ -311,14 +313,28 @@ func (bq *InMemoryBuildQueue) Execute(in *remoteexecution.ExecuteRequest, out re
...
+				label.String("messaging.message_id", in.ActionDigest.Hash),
+			),
+		)
+		desiredState.W3CTraceContext = map[string]string{}
+		util.PropagateInjectMap(ctx, desiredState.W3CTraceContext)
```
EdSchouten, line 331: This would be easier to understand if it could just be written as something like:
```go
desiredState.W3CTraceContext = util.ExtractW3CTraceContextFromContext(ctx)
```
or something like that.

pkg/builder/build_client.go
asuffield: I've guessed a little at what you mean here.

pkg/builder/in_memory_build_queue.go
asuffield: Most of the change is in the other PR, but done.

pkg/builder/in_memory_build_queue.go
asuffield: It can be partially factored out; now done.

I've taken a stab at working platform into this, but it's a bit of a mess. Notably bb_worker never validates its platform config is valid. I've stuffed it in to highlight the problem, let me know if you have any good ideas what to do with this.

go.mod
```diff
@@ -2,29 +2,21 @@ module github.com/buildbarn/bb-remote-execution
 
 go 1.14
 
+replace github.com/buildbarn/bb-storage => github.com/asuffield/bb-storage v0.0.0-20200901154503-006c26285e62
```
asuffield, line 5: Not to be merged, just here to get CI to build.

pkg/builder/in_memory_build_queue.go
asuffield: This is the major outstanding puzzle - we still need to figure out how best to solve this. I'll make another pass later this week and see if I have any better ideas.

EdSchouten:
Fixed in 378ec9d49915d421808d4693568b031c206c668f

Issue #39, opened 2020-02-26
bb-scheduler web view listing sessions


ulrfa:
The bb-scheduler has already a web view listing individual operations/actions. But an additional view listing bazel client invocations, instead of the individual operations/actions they consist of, would give a much better overview to understand what is going on in the cluster.

## Example of a new view listing bazel invocations

Invocation Id | First Action | Priority | Queued Actions | Executing Actions | Finished Actions | Failed Actions | Keywords
-------------|--------------------|-----|----|--------|------|-------|-------------
d252ce12… |2020-02-25 15:44 |  10 |      |           | 1202  |      2 | host=machine3 user=bob
182bfe59…  |2020-02-25 16:13 |   0  | 82 |     28  |  301  |         |host=machine7 user=alice
372bce18… |2020-02-25 16:14 |    0 | 53 |          |          |          |host=machine3 user=bob

#### Notes about each column:
* **Invocation Id**: tool_invocation_id field in REv2 protocol.
* **First Action**: Queue time for first action with this invocation id.
* **Priority**:  E.g. of the first action with this invocation id.
* **Queued Actions**: Number of actions currently waiting in queue with this invocation Id.
* **Executing Actions**: Number of currently executing actions with this invocation Id.
* **Finished Actions**: Number of finished actions with this invocation Id.
* **Failed Actions**: Number of failed actions with this invocation Id.
* **Keywords**: Perhaps derived from information passed via bazel’s new `--remote_exec_header` option.

There should also be:

* a link for listing all individual operations/actions, for a particular Invocation Id.
* a button for aborting all actions with a particual invocation id.

Optionally there could also be a link to a bb-browser web interface for the particular invocation Id, but all builds handled by bb-scheduler might not be using BES and should still be shown in the bb-scheduler view.

EdSchouten:
Hi Ulrik,

Though I think having a page that displays information like this would be awesome to have, we should be careful with adding stuff like that to bb-scheduler itself. The idea right now is that bb-scheduler only displays information that it needs for its own bookkeeping. Aggregating data based on invocation ID, together with counters based on historical execution is something that isn't strictly needed to schedule build actions, right?

That said, it would make sense to add some kind of hooks mechanism to bb-scheduler, so that it could post build action execution events to some helper service (e.g., a more advanced system that combines this information with BES).

ulrfa:
Hi Ed,

[...]

Issue #38, opened 2020-02-26
Schedule actions based on invocation id


ulrfa:
It is often better to finish work instead of letting new work in.

## Example

1.	Bazel client A submits a batch of actions to remote queue.
2.	Then bazel client B submits a huge batch of actions. They are queued waiting to be executed.
3.	Then client A submits a small final set of actions for its session.
4.	Then client B submits many more actions.

#### Current behavior

Actions submitted in step 2 are executed before actions from step 3.

#### Wanted behavior

The actions submitted by client A in step 3, should be executed before queued actions from client B in step 2. Because that would allow the already started client A to finish, and therefore reduce the average latency of completed build sessions.

## Suggestion for bb-scheduler

- Keep track of initial QeueuedTimeStamp for each new bazel invocation (a.k.a. `tool_invocation_id` in REv2). Call it **invocationTimeStamp** and associate it with the `tool_invocation_id`.

- Sort actions in queue based on:
           1. Priority
           2. **InvocationTimeStamp**
           3. QeueuedTimeStamp

- Create some mechanism for discarding information about old `tool_invocation_ids`.

## Discussion

It could be argued that bb-scheduler can’t know if client B has much less remaining work to do than A, and would finish earlier if given a chance. That could be a reason for allowing a limited set of resources also to B, before A is finished. But that might be hard to accomplish without making the queue algorithm much more complex. And the need for such complexity would be reduced if actions can be executed both locally and remotely and taking result from what finishes first, like bazels `--experimental_spawn_scheduler` option.

With other proprietary solutions, we experienced that a client machine could become an bottleneck for the whole cluster, if not balancing actions from several clients, but that should not occur with bazel remote execution, since actions are not queued until after their input files have been uploaded.

An alternative could be to use `correlated_invocations_id` instead of `tool_invocation_id` in REv2.

What do you think?

EdSchouten:
I'm going to close this, as it's unlikely that work is going to be done to satisfy this use case. The reason being:

> It is often better to finish work instead of letting new work in.

This contradicts with what I've observed on my end. People may schedule the weirdest builds that end up hogging resources. Letting the scheduler be fair ensures that a decent amount of resources remain available, even if there are one or more builds that make little forward progress.

Furthermore, with the fairness that we have implemented right now, builds that only require rebuilding a small number of actions can effectively 'overtake' the large builds. I think this is a very desirable property in practice:
- People who make tiny one-line changes want to get a pass in CI quickly.
[...]

PR #72, opened 2020-09-29
Schedule actions based on invocation id, with web view

perwestling:
(implements #38 and #39)
Do group actions based on their invocation ID, to improve flow of invocations.

If having same priority, operations that are queued are sorted
by invocation timestamp before queued timestamp.

Clean up invocation information after a configurable timeframe.

Invocations may share an operation. Keep a list  of connected invocations
in an operation, so that it wont be canceled until all connected invocations
are canceled.

Added invocation count on "Buildbarn Scheduler" view.

Added invocation ID on one operation's view.

Added a view where all invocations can be seen with their current state.
This include a Cancel button (that is enabled unless the invocation has
been canceled). This button will cancel any queued operations, and
also cancel any future operations for this invocation.

Support click on value in many columns in the invocations view.

Add a possibility to configure 'keywords' in bb_scheduler configuration.
This is a list of strings.
E.g.
  keywords: ["host", "user"],
These keywords will be shown in the web view if present in the invocation.
(Bazel client: --remote_exec_header="key=value")

Priority shown in web view. It is possible to specify priority by using 
the --remote_execution_priority=<value> in the bazel call.

Unit test added:
 - that test that invocation timestamp does affect sort order
 - that list of invocations work

Here is an example on how the view looks, in case keywords has been configured (in this case to host and user). The later started invocation has been canceled. 60 seconds (configurable) the last action has completed, the invocation is "moved" to the inactive view. And when Timeout (configured to 60 minutes here) the invocation is cleaned up.
image

### DIFF
--- a/cmd/bb_scheduler/build_queue_state_service.go
+++ b/cmd/bb_scheduler/build_queue_state_service.go
@@ -2,12 +2,15 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"html/template"
 	"log"
 	"net/http"
 	"net/url"
 	"reflect"
+	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	remoteexecution "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
@@ -32,14 +35,15 @@ const (
 var (
 	jsonpbMarshaler         = jsonpb.Marshaler{}
 	jsonpbIndentedMarshaler = jsonpb.Marshaler{Indent: "  "}
+	abbreviateFun           = func(s string) string {
+		if len(s) > 11 {
+			return s[:8] + "..."
+		}
+		return s
+	}
 
 	templateFuncMap = template.FuncMap{
-		"abbreviate": func(s string) string {
-			if len(s) > 11 {
-				return s[:8] + "..."
-			}
-			return s
-		},
+		"abbreviate": abbreviateFun,
 		"action_url": func(browserURL *url.URL, instanceName digest.InstanceName, actionDigest *remoteexecution.Digest) string {
 			d, err := instanceName.NewDigestFromProto(actionDigest)
 			if err != nil {
@@ -73,6 +77,44 @@ var (
 		"to_foreground_color": func(s string) string {
 			return "#" + invertColor(s[:2]) + invertColor(s[2:4]) + invertColor(s[4:6])
 		},
+		"to_timestamp": func(t time.Time) string {
+			localLoc, _ := time.LoadLocation("Local")
+			localDateTime := t.In(localLoc)
+			return localDateTime.Format("15:04:05")
+		},
+		"present_if_value": func(format string, s *string) string {
+			if s == nil {
+				return ""
+			}
+			return fmt.Sprintf(format, abbreviateFun(*s))
+		},
+		"list_keywords": func(keyValueMap map[string]string) string {
+			keys := make([]string, len(keyValueMap))
+
+			i := 0
+			for k := range keyValueMap {
+				keys[i] = k
+				i++
+			}
+			sort.Strings(keys)
+			result := make([]string, i)
+			for idx, key := range keys {
+				result[idx] = fmt.Sprintf("%s=%s", key, keyValueMap[key])
+			}
+			return strings.Join(result, ", ")
+		},
+		"potential_disable": func(canceled bool) string {
+			if canceled {
+				return "disabled"
+			}
+			return ""
+		},
+		"operations_title": func(selection *string) string {
+			if selection == nil {
+				return "All"
+			}
+			return *selection
+		},
 	}
 
 	getBuildQueueStateTemplate = template.Must(template.New("GetBuildQueueState").Funcs(templateFuncMap).Parse(`
@@ -97,6 +139,7 @@ var (
           <th>Instance name</th>
           <th>Platform</th>
           <th>Timeout</th>
+          <th>Invocations</th>
           <th>Queued operations</th>
           <th>Executing workers</th>
           <th>Drains</th>
@@ -109,6 +152,7 @@ var (
           <td>{{.InstanceName}}</td>
           <td>{{$platform}}</td>
           <td>{{with .Timeout}}{{to_duration . $now}}{{else}}∞{{end}}</td>
+          <td><a href="active_invocations?instance_name={{.InstanceName}}">{{.InvocationCount}}</a></td>
           <td><a href="queued_operations?instance_name={{.InstanceName}}&amp;platform={{$platform}}">{{.QueuedOperationsCount}}</a></td>
           <td>
             <a href="workers?instance_name={{.InstanceName}}&amp;platform={{$platform}}&amp;just_executing_workers=true">{{.ExecutingWorkersCount}}</a>
@@ -182,7 +226,7 @@ var (
 <!DOCTYPE html>
 <html>
   <head>
-    <title>All operations</title>
+    <title>{{operations_title .Selection}} operations {{present_if_value "for invocationID %s" .InvocationID}}</title>
     <style>
       html { font-family: sans-serif; }
       table { border-collapse: collapse; }
@@ -192,7 +236,7 @@ var (
     </style>
   </head>
   <body>
-    <h1>All operations</h1>
+    <h1>{{operations_title .Selection}} operations {{present_if_value "for invocationID %s" .InvocationID}}</h1>
     <p>Showing operations [{{.PaginationInfo.StartIndex}}, {{.PaginationInfo.EndIndex}}) of {{.PaginationInfo.TotalEntries}} in total.
       {{with last_element .DetailedOperations}}
         <a href="?start_after_operation={{.Name}}">&gt;&gt;&gt;</a>
@@ -251,6 +295,102 @@ var (
   </body>
 </html>
 `))
+	listActiveInvocationsTemplate = template.Must(template.New("ListActiveInvocations").Funcs(templateFuncMap).Parse(`
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Active bazel invocations</title>
+    <style>
+      html { font-family: sans-serif; }
+      table { border-collapse: collapse; }
+      table, td, th { border: 1px solid black; }
+      td, th { padding-left: 5px; padding-right: 5px; }
+      .text-monospace { font-family: monospace; }
+    </style>
+  </head>
+  <body>
+    {{$instanceName := .InstanceName}}
+    <h1>Recent invocations</h1>
+    <p>Instance name: {{$instanceName}}</p>
+    <p><a href="inactive_invocations?instance_name={{.InstanceName}}">Show inactive invocations</a></p>
+    <table>
+      <thead>
+        <tr>
+          <th>Invocation<br/>ID</th>
+          <th>First<br/>operation</th>
+          <th>Priority</th>
+          <th>Queued<br/>operations</th>
+          <th>Executing<br/>operations</th>
+          <th>Finished<br/>recent<br/>operations</th>
+          <th>Keywords</th>
+          <th>Control</th>
+          <th>Timeout</br>
+        </tr>
+      </thead>
+      {{$browserURL := .BrowserURL}}
+      {{$now := .Now}}
+      {{range .Invocations}}
+        <tr>
+          <td><a href="operations?invocation_id={{.InvocationID}}">{{abbreviate .InvocationID}}</a></td>
+          <td>{{to_timestamp .InvocationTimestamp}}</td>
+          <td>{{.Priority}}</td>
+          <td><a href="operations?invocation_id={{.InvocationID}}&selection=Queued">{{.QueuedOperationsCount}}</a></td>
+          <td><a href="operations?invocation_id={{.InvocationID}}&selection=Executing">{{.ExecutingOperationsCount}}</a></td>
+          <td><a href="operations?invocation_id={{.InvocationID}}&selection=Finished">{{.FinishedOperationsCount}}</a></td>
+          <td>{{list_keywords .Keywords}}</td>
+          <td> <button {{potential_disable .Canceled}} onclick="window.location.href='kill_invocation?name={{.InvocationID}}';" value="">Cancel</button></td>
+          <td>{{with .InvocationTimeout}}{{to_duration . $now}}{{else}}∞{{end}}</td>
+        </tr>
+      {{end}}
+    </table>
+  </body>
+</html>
+`))
+	listInactiveInvocationsTemplate = template.Must(template.New("ListInactiveInvocations").Funcs(templateFuncMap).Parse(`
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Inactive bazel invocations</title>
+    <style>
+      html { font-family: sans-serif; }
+      table { border-collapse: collapse; }
+      table, td, th { border: 1px solid black; }
+      td, th { padding-left: 5px; padding-right: 5px; }
+      .text-monospace { font-family: monospace; }
+    </style>
+  </head>
+  <body>
+    {{$instanceName := .InstanceName}}
+    <h1>Inactive invocations</h1>
+    <p>Instance name: {{$instanceName}}</p>
+    <p><a href="active_invocations?instance_name={{.InstanceName}}">Show active invocations</a></p>
+    <table>
+      <thead>
+        <tr>
+          <th>Invocation<br/>ID</th>
+          <th>First<br/>operation</th>
+          <th>Priority</th>
+          <th>Keywords</th>
+          <th>Control</th>
+          <th>Timeout</br>
+        </tr>
+      </thead>
+      {{$browserURL := .BrowserURL}}
+      {{$now := .Now}}
+      {{range .Invocations}}
+        <tr>
+          <td><a href="operations?invocation_id={{.InvocationID}}">{{abbreviate .InvocationID}}</a></td>
+          <td>{{to_timestamp .InvocationTimestamp}}</td>
+          <td>{{.Priority}}</td>
+          <td>{{list_keywords .Keywords}}</td>
+          <td> <button {{potential_disable .Canceled}} onclick="window.location.href='kill_invocation?name={{.InvocationID}}';" value="">Cancel</button></td>
+          <td>{{with .InvocationTimeout}}{{to_duration . $now}}{{else}}∞{{end}}</td>
+        </tr>
+      {{end}}
+    </table>
+  </body>
+</html>
+`))
 	listQueuedOperationStateTemplate = template.Must(template.New("ListQueuedOperationState").Funcs(templateFuncMap).Parse(`
 <!DOCTYPE html>
 <html>
@@ -279,6 +419,7 @@ var (
       <thead>
         <tr>
           <th>Priority</th>
+          <th>Invocation ID</th>
           <th>Age</th>
           <th>Timeout</th>
           <th>Operation name</th>
@@ -291,6 +432,7 @@ var (
       {{range .QueuedOperations}}
         <tr>
           <td>{{.Priority}}</td>
+          <td>{{abbreviate .InvocationID}}</a></td>
           <td>{{to_duration $now .QueuedTimestamp}}</td>
           <td>{{with .Timeout}}{{to_duration . $now}}{{else}}∞{{end}}</td>
           <td style="background-color: {{to_background_color .Name}}">
@@ -448,7 +590,10 @@ func newBuildQueueStateService(buildQueue builder.BuildQueueStateProvider, clock
 	router.HandleFunc("/", s.handleGetBuildQueueState)
 	router.HandleFunc("/add_drain", s.handleAddDrain)
 	router.HandleFunc("/drains", s.handleListDrainState)
+	router.HandleFunc("/active_invocations", s.handleListActiveInvocations)
+	router.HandleFunc("/inactive_invocations", s.handleListInactiveInvocations)
 	router.HandleFunc("/kill_operation", s.handleKillOperation)
+	router.HandleFunc("/kill_invocation", s.handleKillInvocation)
 	router.HandleFunc("/operation", s.handleGetDetailedOperationState)
 	router.HandleFunc("/operations", s.handleListDetailedOperationState)
 	router.HandleFunc("/queued_operations", s.handleListQueuedOperationState)
@@ -478,6 +623,15 @@ func (s *buildQueueStateService) handleKillOperation(w http.ResponseWriter, req 
 	http.Redirect(w, req, req.Header.Get("Referer"), http.StatusSeeOther)
 }
 
+func (s *buildQueueStateService) handleKillInvocation(w http.ResponseWriter, req *http.Request) {
+	req.ParseForm()
+	if !s.buildQueue.KillInvocation(req.FormValue("name")) {
+		http.Error(w, "Invocation not found: "+req.FormValue("name"), http.StatusNotFound)
+		return
+	}
+	http.Redirect(w, req, req.Header.Get("Referer"), http.StatusSeeOther)
+}
+
 func (s *buildQueueStateService) handleGetDetailedOperationState(w http.ResponseWriter, req *http.Request) {
 	detailedOperation, ok := s.buildQueue.GetDetailedOperationState(req.URL.Query().Get("name"))
 	if !ok {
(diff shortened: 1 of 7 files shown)


erenon:
Assuming a buildbarn plant of 100 workers, 10 developers using the plant, everyone with the same default priority. Wouldn't this change greatly increase the standard deviation of waiting time? The 10 devs sending in their invocations will be queued up, instead of having their invocations equally scheduled, 10 workers for each.

ulrfa:
> Assuming a buildbarn plant of 100 workers, 10 developers using the plant, everyone with the same default priority. Wouldn't this change greatly increase the standard deviation of waiting time? The 10 devs sending in their invocations will be queued up, instead of having their invocations equally scheduled, 10 workers for each.

Thanks for feedback @erenon 

The idea is: Better to finish one, instead of having all of them waiting.

See more elaborate discussion in https://github.com/buildbarn/bb-remote-execution/issues/38

Do you think that make sense?

erenon:
> The actions submitted by client A in step 3, should be executed before queued actions from client B in step 2. Because that would allow the already started client A to finish, and therefore reduce the average latency of completed build sessions

If the action set size of A and B are the same, this does not reduce the average latency, only reduces the min latency, and increases the max latency, keeping the average roughly the same.
If the first set is much smaller, avg decreases.
If the first set is much bigger, avg increases.

For the scheduling algorithm, that picks an action from the queue, wouldn't it be better (i.e: more fair) to round-robin between the active invocation ids?

ulrfa:
> this does not reduce the average latency, only reduces the min latency, and increases the max latency

What do you mean with that @erenon? Do we interpret _"average latency of completed build sessions_" differently?

Example:
```
First invocation scheduling: AAAABBBB                                            
Round-robin scheduling:      ABABABAB
```
Session B finish after 8 time slots with both schedulers.

But the first scheduler finish session A after 4 time slots, instead of 7 with round-robin.

The average is decreased since:  4+8  <  7+8

erenon:
I think we use the same definition: latency is the time elapsed between the first action received and the completion of the full invocation.

In your example, the action set sizes are the same, there's no big difference between the two scheduling algorithms.

Consider a different example: A is large, B is small (e.g: A changes a common header file, B changes a single source file)

```
Input: ABAAAAAAAA
FIFO: AAAAAAAAAAB
Round robin: ABAAAAAAAA
```

Fifo avg latency is (10+11)/2=10.5, rr avg latency is (11+2)/2=6.5

ulrfa:
Example with equal action set sizes for all invocations, but scaled up slightly:

```
                       1           2           3           4           5 
             12345 67890 12345 67890 12345 67890 12345 67890 12345 67890

FIFO:        AAAAA AAAAA BBBBB BBBBB CCCCC CCCCC DDDDD DDDDD EEEEE EEEEE
Round Robin: ABCDE ABCDE ABCDE ABCDE ABCDE ABCDE ABCDE ABCDE ABCDE ABCDE

```
Average latency:
-   FIFO: (10 + 20 + 30 + 40 + 50) / 5 = 30
-   Round Robin: (46 + 47 + 48 + 49 + 50) / 5 = 48

Average latency for Round Robin is (48-30)/30 = 60% slower than invocation based FIFO. I would say that is a big difference.

And the difference could be even larger if invocations started earlier, and therefore completed more actions, on average has fewer remaining actions.

And in extreme overload conditions, it is often better that some invocations complete, instead of all of them timing out  and being aborted after executed half-way.

But you have a good point about invocations only consisting of a few actions. It is unfortunate if they have to wait  for much larger invocations. Perhaps the invocation based FIFO algorithm could be a start, and then consider making the algorithm more advanced as a second step? (E.g. allowing the other invocations x concurrent action(s)). It could also be that the need for such extra complexity is reduced if actions can be executed both locally and remotely speculatively and taking result from what finishes first, like bazels --experimental_spawn_scheduler option. Because then the invocations with just a few actions could finish via local execution.

erenon:
I see your point. It all comes down to the kind of load we are expecting. It occurred to me, that in the past, computers executed punchcard programs in batches - those jobs took a long time. Today, no interactive desktop environment would use FIFO scheduling.
I'm not a scheduling expert, the related Wikipedia section seems to be: https://en.wikipedia.org/wiki/Scheduling_(computing)#Scheduling_disciplines

If we don't expect that there's a scheduler best for all load models, probably we can aim for a configurable scheduler instead, with yours the first (and so far only) one.

> actions can be executed both locally and remotely speculatively
Only if local is capable to execute - not always the case (e.g: cross compiling)

ulrfa:
@EdSchouten do you think you will have time to look into this pull request?

pkg/builder/in_memory_build_queue.go
```diff
@@ -238,7 +320,7 @@ func (bq *InMemoryBuildQueue) GetCapabilities(ctx context.Context, in *remoteexe
 // blocks until the action is completed.
 func (bq *InMemoryBuildQueue) Execute(in *remoteexecution.ExecuteRequest, out remoteexecution.Execution_ExecuteServer) error {
 	// Fetch the action and command corresponding to the execute
-	// request. Ideally, a scheduler is be oblivious of what these
+	// request. Ideally, a scheduler oblivious of what these
```
EdSchouten, line 323: s/is be/is?

pkg/builder/in_memory_build_queue.go
```diff
@@ -1091,13 +1350,26 @@ func (h queuedOperationsHeap) Len() int {
...
-	return h[i].priority < h[j].priority || (h[i].priority == h[j].priority &&
-		(ti.Seconds < tj.Seconds || (ti.Seconds == tj.Seconds &&
-			ti.Nanos < tj.Nanos)))
+func (h queuedOperationsHeap) Less(x, y int) bool {
+	// Lexicographic order on priority, invocation timestamp and queued timestamp.
+	// The invocation timestamp used is the oldest timestamp for connected invocations.
+	qx := h[x].operation.desiredState.QueuedTimestamp
+	qy := h[y].operation.desiredState.QueuedTimestamp
+
+	return h[x].priority < h[y].priority ||
```
EdSchouten, line 1359: I think the readability of this function could be improved if it was written as follows:
```go
if vx, vy := h[x].priority, h[y].priority; vx < vy {
    return true
} else if vx > vy {
    return false
}
if vx, vy := h[x].operation.invocations[0].startTime, h[y].operation.invocations[0].startTime; vx < vy {
    return true
} else if vx > vy {
    return false
}
...
```
This allows us to keep on adding more conditions without blowing up expression size.

That said, are we happy with the selection logic that is used here? It looks like you’re picking the start time of the build to determine which action gets preferred over the other one, right? Doesn’t this mean that long-running builds that have been running for some time will cause starvation for short-running builds that have only been started recently?

My suggestion would be to basically replace queuedOperationsHeap by a different data structure: a heap of heaps.

- The top-level heap would contain every Bazel invocation for which one or more build actions are queued for that platform. This is a min-heap that is ordered by the number of actions that this invocation has running on that platform right now.
- The second-level heap would contain every operation associated with that Bazel invocation. This is a min-heap that is ordered by priority and queueing time.

By using an approach like this, every Bazel client will get its fair share of workers. The scheduler will continuously allow the Bazel client that runs the least number of build actions to run more actions.

EdSchouten (commented):
Hi Per,

Sorry for letting you wait! The last couple of weeks have been far too busy on my end. Things should start cooling down now, so I should have enough time to review your changes.

Ed

EdSchouten:
> The idea is: Better to finish one, instead of having all of them waiting.

Though that sounds like a good idea at the surface, there is also a big risk to this. What if someone makes a code change that causes a large number of tests to hang, and runs that with `-j<high>`? That would cause other builds to get stalled for a long amount of time.

In many cases, people use Buildbarn for shared setups, where a large number of people send build actions concurrently. For workloads like those, it is far safer to not let one build dominate all the workers.

cmd/bb_scheduler/build_queue_state_service.go
```diff
@@ -182,7 +226,7 @@ var (
 <!DOCTYPE html>
 <html>
   <head>
-    <title>All operations</title>
+    <title>{{operations_title .Selection}} operations {{present_if_value "for invocationID %s" .InvocationID}}</title>
```
EdSchouten, line 229: The `present_if_value` macro doesn't really carry its own weight. What's the advantage of this over `{{with .InvocationID}} for invocation {{.}}{{end}}`?

cmd/bb_scheduler/build_queue_state_service.go
```diff
@@ -73,6 +77,44 @@ var (
...
+			localDateTime := t.In(localLoc)
+			return localDateTime.Format("15:04:05")
+		},
+		"present_if_value": func(format string, s *string) string {
+			if s == nil {
+				return ""
+			}
+			return fmt.Sprintf(format, abbreviateFun(*s))
+		},
+		"list_keywords": func(keyValueMap map[string]string) string {
```
EdSchouten, line 91: Am I correct that the key-value pairs provided by Bazel are never used as if they were a map? It's just an opaque set of values that we get from Bazel that are displayed in the web UI. Instead of letting the build queue state service expose it as a map, I guess it would make more sense to just serve a slice of key-value pairs that are sorted up front?

cmd/bb_scheduler/build_queue_state_service.go
```diff
@@ -250,6 +294,102 @@ var (
...
+      {{range .Invocations}}
+        <tr>
+          <td><a href="operations?invocation_id={{.InvocationID}}">{{abbreviate .InvocationID}}</a></td>
+          <td>{{to_timestamp .InvocationTimestamp}}</td>
+          <td>{{.Priority}}</td>
+          <td><a href="operations?invocation_id={{.InvocationID}}&selection=Queued">{{.QueuedOperationsCount}}</a></td>
+          <td><a href="operations?invocation_id={{.InvocationID}}&selection=Executing">{{.ExecutingOperationsCount}}</a></td>
+          <td><a href="operations?invocation_id={{.InvocationID}}&selection=Finished">{{.FinishedOperationsCount}}</a></td>
+          <td>{{list_keywords .Keywords}}</td>
+          <td> <button {{potential_disable .Canceled}} onclick="window.location.href='kill_invocation?name={{.InvocationID}}';" value="">Cancel</button></td>
```
EdSchouten, line 341: The same holds for potential_disable: `{{if .Canceled}}disabled{{end}}` seems to be as long as calling into this helper function.

pkg/builder/in_memory_build_queue.go
ulrfa: I like the idea about ordering bazel invocations, rather than actions, in a top-level heap.

But shouldn’t the priority field be more significant? Not having priority present at all in the top-level heap would make priority almost useless. Right?

And if there are several invocations with equal number of actions running, then the oldest invocation should be scheduled first, to avoid starvation. Right?

It seems that there is no single algorithm that fit all use cases and expected load models. That could be a good reason for making it configurable. Would the following give enough flexibility for most use cases?

Top-level heap of bazel invocations ordered by:
1. priority + max(0, (numberOfActionsRunningForTheInvocation - X))
2. invocation start time

where X would be an integer set in bb-schedulers configuration file. Setting X=0 would share capacity equal between invocations, similar to what @EdSchouten suggested. A very high X value would schedule oldest invocation, as this PR is doing. Configuring X to something in between would be a compromise between the two extremes.

I’m not sure of a good name for ‘X’. Maybe ReservedCapacityPerInvocation?

Adding the priority and numberOfActionsRunningForTheInvocation integers together gives flexibility for different use cases, since a priority close to the default 0 would influence the scheduling slightly, while a priority value like -100000 would be very significant.

What do you think @EdSchouten?

EdSchouten: I think it depends on how 'priority' should be interpreted. It can be interpreted in two different ways:
- There is the concept of high priority vs low priority builds. If two invocations of Bazel are running concurrently, which one should get more workers than the other?
- There is the concept of action-level priority. Given a set of actions being scheduled as part of a single build, which action can best be run first to ensure that the total duration of the build stays close to the critical path?

In case the idea behind REv2's priority field was to cater to that second use case, then the approach that I laid out is exactly what we'd want: treat all builds equal, but let the priority field completely dominate what gets scheduled as part of that build.

In case it's the former, then things become hairy. Especially considering the fact that it's not as if a Bazel invocation has a single priority. It's composed of build actions that *each* have their own priority. It's basically a two-dimensional thing, and we need to turn it into something one-dimensional. Which means I agree with your statement:
> It seems that there is no single algorithm that fit all use cases and expected load models.

That said, if we were to simply make the assumption that Bazel always sends the same priority value for all of its build actions, then I think we can come up with a strategy that requires a bit less magic than what you laid out.

What if the default policy (in case priority == 0 for all builds) is that all builds get exactly the same set of workers (so n/k, where n is the number of workers, k is the number of Bazel builds going on at that point in time). Then, the priority value can be used to influence a weight factor.

The naïve approach would be to use the priority directly as the weight factor through multiplication, but that's annoying. You have to deal with priority being non-positive. This is why you came up with this expression:

> priority + max(0, (numberOfActionsRunningForTheInvocation - X))

Instead of using that, what are your thoughts on using an exponential weight? An awesome property of exponentiation is that the exponent can be an arbitrary number, and the output is always positive (assuming a positive base). So if you set priority to zero, your build gets weight `b^0 = 1`. If you pick priority=1, you get `b` times as many (as `b^1 = b`). It doesn't require any explicitly configured bounds on the scheduler.

Now the question is what `b` should be. Let me propose a value: `b = 2^(1/100) =~ 1.00696`. This means that if you set priority=100, you get twice as many workers as another build running with priority=0. If you set priority=200, you get four times as many. That should give people more than enough granularity.

ulrfa: I don’t know about BuildStrem, RECC, etc, but I have not seen support in bazel for priorities per action, only per invocation. Maybe I missed something? If not, is it good enough taking priority from first seen action, as in this PR?

If we would like to honor the granularity in REv2, we could consider letting each invocation in top-level heap, inherit the priority of next queued action, from its associated second-level heap. And/or have a configuration option for ignoring priority in top-level heap?

> This is why you came up with this expression

Yes, you are reading my mind :smiley:

Your idea with `b^priority` seems better than my attempt adding integers.  Do you imagine the weight factor to be used like the following?

`(b^priority) * (1 + numberOfActionsRunningForTheInvocation)`

And if also including a configurable X parameter, could the top-level heap be ordered by the following?

1.	`(b^priority) * (1 + max(0, (numberOfActionsRunningForTheInvocation – X)))`
2.	`invocation start time`

EdSchouten: I'd say that the top level uses a min-heap, sorted by the expression:
```
number_of_actions_running_for_the_invocation / b^priority
```

ulrfa: > I'd say that the top level uses a min-heap, sorted by the expression:
> 
> ```
> number_of_actions_running_for_the_invocation / b^priority
> ```

-100 in REv2 priority field means higher priority than 100. Right?

Higher: `b^-100 =~ 0.5`
Lower: `b^100 =~ 2`

So for a min-heap where we want the higher priority to get the smallest value, we need to multiply, not divide. Right?

`number_of_actions_running_for_the_invocation * b^priority`

And when there are several invocations with zero actions running, their resulting product would all be 0, **regardless of their priority**. To assure they are started in priority order, we can calculate:

`(1 + number_of_actions_running_for_the_invocation) * b^priority`

> Surely we can take the start time into account as well, but I don't think that's really necessary. I don't think that can lead to sustainable starvation of any sort, right? Every time a certain invocation gets an operation assigned to a worker, it'll become harder for it to get more operations assigned, due to number_of_actions_running_for_the_invocation increasing.

That is true when cluster capacity is significantly higher than number of invocations, but not if there are many invocations waiting without any of their actions running.

And if having a configurable X constant in the formula, for my use case, then that secondary ordering by invocation start time would be essential.

EdSchouten: You're 100% correct!

EdSchouten:
Hi there,

I wanted to experiment with the approach that @ulrfa and I discussed related to using exponentiation as a weight factor for prioritising builds. This is what I came up with:

https://github.com/buildbarn/bb-remote-execution/compare/eschouten/20201028-scheduler-grouping

I've done some experimenting on my end, and I'm able to get pretty much perfect fairness for builds with a change like this applied. Please review!

@perwestling This patch should also be able to cater to your needs. Just make sure that you launch builds with the priority set to `$(date +%s)` and you'll see that older builds get a higher priority than newer ones.

ulrfa:
> https://github.com/buildbarn/bb-remote-execution/compare/eschouten/20201028-scheduler-grouping
> 
> Please review!

Thanks @EdSchouten, I will try to find time to review during the weekend.

ulrfa:
> @perwestling This patch should also be able to cater to your needs. Just make sure that you launch builds with the priority set to `$(date +%s)` and you'll see that older builds get a higher priority than newer ones.

I tried to elaborate about my use case in comments to https://github.com/buildbarn/bb-remote-execution/commit/7faeb002b216505bc61b14840096bd471297df41

Some additional comments:

In our environment, a new git commit often triggers several new invocations in CI, almost the same time. Using `$(date +%s)` as priority would give them almost same priority even if 10 seconds between. Millisecond precision would overflow the uint32 REv2 priority field.

And when needing also conceptual priority levels at the same time, it would be complex to merge both the conceptual priority levels and the timestamp into a single uint32. Users seeing such merged value in web view won't understand the conceptual priority level.

And for monitoring, I would like to plot the queue length (or similar) grouped by priority. E.g. queues of low priority batch invocations are expected, but capacity should be increased if continuously having long queues of high priority interactive invocations. It would be harder to create Prometheus counters for queue length, grouped per priority, if the priority values would be calculated from timestamps and not only consist of an enumerated set of more fixed priorities.

perwestling:
@EdSchouten A couple of months ago I was working on this one, supported by Ulrik Falklöf (@ulrfa); we were working in the same team. But at the end of September I moved to another project, and have not had time to look into this again. 
So you can take the feedback from Ulrik as if it came from me. 

How do we progress with this issue? Will you work more on https://github.com/buildbarn/bb-remote-execution/compare/eschouten/20201028-scheduler-grouping, or should we try to incorporate it into this commit?

EdSchouten:
I've just merged that branch of mine into master. :tada:

> In our environment, a new git commit often triggers several new invocations in CI, almost the same time. Using `$(date +%s)` as priority would give them almost same priority even if 10 seconds between. Millisecond precision would overflow the uint32 REv2 priority field.
> 
> And when needing also conceptual priority levels at the same time, it would be complex to merge both the conceptual priority levels and the timestamp into a single uint32. Users seeing such merged value in web view won't understand the conceptual priority level.

That's a fair point. We'll need to think of some other mechanism then. That said, I'm a bit hesitant towards supporting this in the absolute sense that you are considering, for the reason that it's also a massive footgun. One CI run with a change applied that makes all build actions hang (note: non-test targets don't specify any timeout) and the entire CI pipeline will stall because of it. I don't think that's a good policy.

> And for monitoring, I would like to plot the queue length (or similar) grouped by priority. E.g. queues of low priority batch invocations are expected, but capacity should be increased if continuously having long queues of high priority interactive invocations. It would be harder to create Prometheus counters for queue length, grouped per priority, if the priority values would be calculated from timestamps and not only consist of an enumerated set of more fixed priorities.

Exactly. Prometheus doesn't really like it if metrics have arbitrary cardinality. Though I agree it would be nice to expose metrics by priority, I don't think that's feasible due to it being a user-provided value.

ulrfa:
> That's a fair point. We'll need to think of some other mechanism then. That said, I'm a bit hesitant towards supporting this in the absolute sense that you are considering, for the reason that it's also a massive footgun. One CI run with a change applied that makes all build actions hang (note: non-test targets don't specify any timeout) and the entire CI pipeline will stall because of it. I don't think that's a good policy.

The new "fair share" scheduler is not immune to this either. @EdSchouten, is there no other more reliable way to protect against hanging actions? Which could also manage the situation: hanging actions that got all resources, **during a time with no other competing invocations (or only lower priority ones)**, and then continue hanging while more invocations arrive.

For me, the purpose of remote execution is to provide massive amount of resources, just when needed. Limiting the resources due to suspicions that they will not be released, seems unfortunate.

@EdSchouten, have you seen my 3 comments in in_memory_build_queue.go in https://github.com/buildbarn/bb-remote-execution/commit/7faeb002b216505bc61b14840096bd471297df41? They are not easy to spot in the large file. Via  configuration.ReservedCapacityPerInvocation users would have the choice of leaving it at default 0, or increasing it to e.g. 25% or 100% of the capacity, depending on how they assess the risk.

> Exactly. Prometheus doesn't really like it if metrics have arbitrary cardinality. Though I agree it would be nice to expose metrics by priority, I don't think that's feasible due to it being a user-provided value.

In its simplest form, queue metrics could be grouped in three buckets:

- Priority < 0
- Priority == 0
- Priority > 0

Other bucket boundaries could be set in bb-scheduler config file, for adapting to the priority levels used in an organization. Do you think such buckets would make sense?

EdSchouten:
> The new "fair share" scheduler is not immune to this either. @EdSchouten, is there no other more reliable way to protect against hanging actions? Which could also manage the situation: hanging actions that got all resources, **during a time with no other competing invocations (or only lower priority ones)**, and then continue hanging while more invocations arrive.

Yeah, but wouldn't that be better solved by adding true support for preemption to the scheduler? Adding that is to a fair extent orthogonal to the scheduler changes I made earlier.

> For me, the purpose of remote execution is to provide massive amount of resources, just when needed. Limiting the resources due to suspicions that they will not be released, seems unfortunate.

I guess everybody has a different use case for their build cluster. The use case that I'm focusing on with Buildbarn is to be able to offer a multi-tenant cluster that is both used for CI builds and ones performed by people at desk.

> @EdSchouten, have you seen my 3 comments in in_memory_build_queue.go in [7faeb00](https://github.com/buildbarn/bb-remote-execution/commit/7faeb002b216505bc61b14840096bd471297df41)? They are not easy to spot in the large file. Via configuration.ReservedCapacityPerInvocation users would have the choice of leaving it at default 0, or increasing it to e.g. 25% or 100% of the capacity, depending on how they assess the risk.

Something like this wouldn't be needed if we had preemption.

> > Exactly. Prometheus doesn't really like it if metrics have arbitrary cardinality. Though I agree it would be nice to expose metrics by priority, I don't think that's feasible due to it being a user-provided value.
> 
> In its simplest form, queue metrics could be grouped in three buckets:
> 
> * Priority < 0
> * Priority == 0
> * Priority > 0
> 
> Other bucket boundaries could be set in bb-scheduler config file, for adapting to the priority levels used in an organization. Do you think such buckets would make sense?

I don't know. That seems a bit arbitrary. What if people use priorities, 100, 200, 300, 400? Maybe it's better if access to data in such a granular way is done through stuff like OpenTracing/OpenCensus.

ulrfa:
> > @EdSchouten, have you seen my 3 comments in in_memory_build_queue.go in [7faeb00](https://github.com/buildbarn/bb-remote-execution/commit/7faeb002b216505bc61b14840096bd471297df41)? They are not easy to spot in the large file. Via configuration.ReservedCapacityPerInvocation users would have the choice of leaving it at default 0, or increasing it to e.g. 25% or 100% of the capacity, depending on how they assess the risk.
> 
> Something like this wouldn't be needed if we had preemption.

Perhaps I caused confusion by copy&paste the wrong parameter. I meant `configuration.ReservedCapacityForEarlierInvocation`. That allows giving some capacity for free to earlier invocations, before they have to compete with later invocations of the same priority. In order to favor completing already started earlier invocations, rather than spreading capacity even.

How could support for preemption solve that use case?

> Yeah, but wouldn't that be better solved by adding true support for preemption to the scheduler? Adding that is to a fair extent orthogonal to the scheduler changes I made earlier.

Regarding preemption, you said my proposal was a ‘massive footgun’ and if that is the major objection, I wanted to highlight that the same problem can occur also without my proposal. Yes, it is to a fair extent orthogonal.

> That seems a bit arbitrary. What if people use priorities, 100, 200, 300, 400?

Then buckets with those boundaries could be configured in bb-scheduler config file. Perhaps the idea about metrics per priority, is best tracked as a separate ticket.

> Maybe it's better if access to data in such a granular way is done through stuff like OpenTracing/OpenCensus.

I don't have experience of OpenTracing/OpenCensus.

I understand you want the scheduler to only track the bare minimum amount of information that is needed to make scheduling decisions. @EdSchouten, **is it completely out of the question to make exception and track also invocation start time and a set of keywords per session?** In order to support more features in the web admin interface, and scheduling algorithm extensions such as `configuration.ReservedCapacityForEarlierInvocation`. Similar to the argv0 string in the task type. If not acceptable, the only remaining option for my company might be to consider managing our own buildbarn patches, but I would really want to avoid that.

EdSchouten:
Closing due to inactivity.

PR #73, opened 2020-12-03
Add support for compiling on 32-bit systems

Gormo:
A few explicit casts are needed in order to get the
complete buildbarn code to compile on 32-bit systems.

This makes it possible to e.g compile and execute buildbarn
on raspberry pi:s running on 32-bit kernels.

### DIFF
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -77,7 +77,7 @@ func main() {
 			log.Fatal("Failed to open file pool block device: ", err)
 		}
 		if sectorCount > math.MaxUint32 {
-			log.Fatal("File pool block device has %d sectors, while only %d may be addressed", sectorCount, math.MaxUint32)
+			log.Fatal("File pool block device has %d sectors, while only %d may be addressed", sectorCount, uint64(math.MaxUint32))
 		}
 		filePool = re_filesystem.NewBlockDeviceBackedFilePool(
 			blockDevice,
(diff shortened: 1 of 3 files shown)


cmd/bb_worker/main.go
```diff
@@ -77,7 +77,7 @@ func main() {
 			log.Fatal("Failed to open file pool block device: ", err)
 		}
 		if sectorCount > math.MaxUint32 {
-			log.Fatal("File pool block device has %d sectors, while only %d may be addressed", sectorCount, math.MaxUint32)
+			log.Fatal("File pool block device has %d sectors, while only %d may be addressed", sectorCount, uint64(math.MaxUint32))
```
EdSchouten, line 80: Ah, this should have been a log.Fatalf(). Could you fix this while you're at it?

What's the error you're getting here? That max.MaxUint32 doesn't fit in a plain int? Could you maybe cast it to a uint32 instead? That should be big enough to fit this constant.

EdSchouten (approved):
Thanks! Just some minor tweaks and this is good to land.

cmd/bb_worker/main.go
EdSchouten: You know what? I'll make these changes for you. My plan is to enable CI for linux-386 at the same time, so that support for 32-bit platforms should remain stable going forward.

Gormo: Ok, I'll change it to Fatalf! The error that I'm getting is: 
`
main.go:80:14: constant 4294967295 overflows int
`

PR #79, opened 2021-03-07
Don't check OutputFiles and OutputDirs if OutputPaths is set

peterebden:
According to the API, 
```
If `output_paths` is used, `output_files` and `output_directories` will be ignored!
```
We were setting all three in plz for compatibility; I'm going to stop sending the old ones (in https://github.com/thought-machine/please/pull/1624) which will fix our use case, but thought you might appreciate this in case anyone else does the same thing while they're migrating!


EdSchouten:
Hey! Change looks good, but don’t forget to patch up the tests.

peterebden:
Yep sorry, took me a little while to get set up with Bazel etc & to figure out why some of them were failing, but should be good now.

EdSchouten:
No worries. Thanks for submitting this change!

PR #82, opened 2021-03-29
Add environment variables to worker config

jlaxson:
Allows adding env vars to those given by the action.

This is needed for the xcode toolchain, because the clang wrapper used by bazel depends on two variables, DEVELOPER_DIR and SDKROOT, to locate the correct toolchain to use.  These are set beneath the main action layer (they aren't reported for instance when you do `bazel build ... -s`).  I believe the behavior must be implemented somewhere in buildbarn because:
1. `--action_env` rather comically doesn't actually apply to most actions (bazelbuild/bazel#3320).  Notably in my case it wasn't passed to cc linking actions.
2. The runner (correctly, in my opinion) does not pass on its surrounding environment to the actions it spawns, so I can't set these before running bb_runner.

I put the specification in bb_worker's RunnerConfiguration because it was close in proximity to the platform config, which the variables are likely to relate to (e.g. add variables to support a particular toolchain advertised by Platform).  But there are probably other places they could go too.  As it exists now, these variables are overridden by anything specified in the action's env, but I could see potential use cases for wanting to specifically prepend/append to `PATH`.

I think I saw a conversation in slack that was not enthusiastic about carrying these inside bb, but didn't really see another viable option (and this is roughly what I recall buildfarm recommending)

### DIFF
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -265,11 +265,7 @@ func main() {
 						log.Fatal("Failed to marshal worker ID: ", err)
 					}
 
-					env := map[string]string{}
-					for k, v := range runnerConfiguration.EnvironmentVariables {
-						env[k] = v
-					}
-
+					env := runnerConfiguration.EnvironmentVariables
 					buildExecutor := builder.NewMetricsBuildExecutor(
 						builder.NewFilePoolStatsBuildExecutor(
 							builder.NewTimestampedBuildExecutor(
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -265,6 +265,11 @@ func main() {
 						log.Fatal("Failed to marshal worker ID: ", err)
 					}
 
+					env := map[string]string{}
+					for k, v := range runnerConfiguration.EnvironmentVariables {
+						env[k] = v
+					}
+
 					buildExecutor := builder.NewMetricsBuildExecutor(
 						builder.NewFilePoolStatsBuildExecutor(
 							builder.NewTimestampedBuildExecutor(
@@ -277,7 +282,8 @@ func main() {
 										defaultExecutionTimeout,
 										maximumExecutionTimeout,
 										inputRootCharacterDevices,
-										int(configuration.MaximumMessageSizeBytes)),
+										int(configuration.MaximumMessageSizeBytes),
+										env),
 									contentAddressableStorageFlusher),
 								clock.SystemClock,
 								string(workerName))))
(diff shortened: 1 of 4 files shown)


cmd/bb_worker/main.go
```diff
@@ -265,6 +265,11 @@ func main() {
 						log.Fatal("Failed to marshal worker ID: ", err)
 					}
 
+					env := map[string]string{}
```
EdSchouten, line 268: Is this loop really needed? Couldn’t you just pass the original map along?

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -153,4 +153,8 @@ message RunnerConfiguration {
   // auxiliary_metadata.
   map<string, buildbarn.resourceusage.MonetaryResourceUsage.Expense>
       costs_per_second = 10;
+
+  // Additional environment variables to set inside the runner.  These are
```
EdSchouten, line 157: Maybe add a comment there that even though this option is available, the use of action_env is preferred?

cmd/bb_worker/main.go
jlaxson: Yeah I wasn't sure, if the main map is copied by value, then skip the loop and initialize env as a copy of runnerConfiguration.EnvironmentVariables?  Not quite a go sage yet.

pkg/proto/configuration/bb_worker/bb_worker.proto
jlaxson: Sure

jlaxson:
If this is acceptable in principle I'll work out updating the tests

cmd/bb_worker/main.go
EdSchouten: It should be safe to just pass it along. We don't mutate any of the structs we parse from the config.

EdSchouten (commented):
> If this is acceptable in principle I'll work out updating the tests

Sure! No need to add a fully separate test. Be sure to modify one or two of the existing ones.

pkg/proto/configuration/bb_worker/bb_worker.proto
EdSchouten: Great! Be sure to also put the link to the Bazel issue report in there, so people are aware of the limitations of action_env.

cmd/bb_worker/main.go
jlaxson: Yes.  I was confusing it for the use inside the Executor where we do need to make a copy to modify it with the action-specified environment

pkg/builder/local_build_executor_test.go
```diff
@@ -584,7 +585,11 @@ func TestLocalBuildExecutorSuccess(t *testing.T) {
 	inputRootCharacterDevices := map[path.Component]int{
 		path.MustNewComponent("null"): 259,
 	}
-	localBuildExecutor := builder.NewLocalBuildExecutor(contentAddressableStorage, buildDirectoryCreator, runner, clock, time.Hour, time.Hour, inputRootCharacterDevices, 10000)
+	environmentVars := map[string]string{
```
EdSchouten, line 588: Great! Thanks for testing both the behaviour of plain assignment and overwriting an existing one.

EdSchouten (approved):

PR #83, opened 2021-03-31
HardlinkFetcher can use Clonefile

jlaxson:
In concert with buildbarn/bb-storage#116, runner config now specifies use of hardlinks or clonefile for populating build directories.


cmd/bb_worker/main.go
```diff
@@ -150,12 +150,22 @@ func main() {
 			if err != nil {
 				log.Fatal("Failed to create eviction set for cache directory: ", err)
 			}
+			var linkHelper cas.LinkHelper
```
EdSchouten, line 153: What are your thoughts on changing the naming to something like this?
- HardlinkingFileFetcher -> CachingFileFetcher
- LinkHelper -> FileInstaller
  - LinkHelperHardlink -> HardlinkingFileInstaller
  - LinkHelperClonefile -> CloningFileInstaller

I don't think HardlinkingFileFetcher is accurate anymore, now that it can also be used to clone files.

pkg/cas/BUILD.bazel
```diff
@@ -11,6 +11,9 @@ go_library(
         "directory_walker.go",
         "file_fetcher.go",
         "hardlinking_file_fetcher.go",
+        "link_helper.go",
```
EdSchouten, line 14: These source files are missing?

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -110,6 +111,15 @@ message NativeBuildDirectoryConfiguration {
   // LEAST_RECENTLY_USED can, assuming the cache size is sufficient.
   buildbarn.configuration.eviction.CacheReplacementPolicy
       cache_replacement_policy = 5;
+
+  NativeBuildDirectoryLinkType link_type = 6;
```
EdSchouten, line 115: I'm fine with just using a boolean here: `use_clonefile`. We can always extend it to have a `oneof` by the time we add even more mechanisms ;-)

pkg/cas/hardlinking_file_fetcher.go
```diff
@@ -78,7 +82,7 @@ func (ff *hardlinkingFileFetcher) GetFile(ctx context.Context, blobDigest digest
 		ff.evictionSet.Touch(key)
```
EdSchouten, line 82: About 10 lines up from here, you can see that we generate a key for every object: `${sha}-${size}${isExecutable}`.

When doing cloning, we don't need to take `isExecutable` into account. Every clone can have its own permissions, meaning the cache doesn't need to hold up to two copies of every file. Do we want to make use of that property, or do we find it unlikely that this is going to reduce disk space in any measurable way? If it's the latter, can we add a comment up there that states this? Thanks!

pkg/cas/hardlinking_file_fetcher.go
```diff
@@ -121,7 +125,7 @@ func (ff *hardlinkingFileFetcher) GetFile(ctx context.Context, blobDigest digest
 	} else if wasMissing {
 		// Even though the file is part of our bookkeeping, we
 		// observed it didn't exist. Repair this inconsistency.
-		if err := directory.Link(name, ff.cacheDirectory, path.MustNewComponent(key)); err != nil && !os.IsExist(err) {
+		if err := ff.linkHelper.Link(directory, name, ff.cacheDirectory, path.MustNewComponent(key)); err != nil && !os.IsExist(err) {
```
EdSchouten, line 128: Does `Clonefile()` preserve the file modification time, or does the cloned instance get its modification time updated? In case it's the latter, do we want to add an additional call to reset it back to Jan 1st, 2000? Some build actions (mainly ones using rules_foreign_cc) get annoyed if we don't ensure all input files have the same modification time.

cmd/bb_worker/main.go
```diff
@@ -150,12 +150,22 @@ func main() {
...
+			switch nativeConfiguration.LinkType.LinkType.(type) {
+			case *bb_worker.NativeBuildDirectoryLinkType_Hardlink:
+				linkHelper = cas.NewLinkHelperHardlink()
```
EdSchouten, line 156: For unparameterized types like these, we often follow a pattern where we just provide a singleton instance. See 'InMemoryFilePool' for an example. Could we also use a construct like that over here?

pkg/cas/hardlinking_file_fetcher.go
jlaxson: An unscientific check using `cp -c` shows the mod time preserved (whereas normal cp does not).

Although I have decided that the correct answer when using rules_foreign_cc is "burn it with fire"

cmd/bb_worker/main.go
jlaxson: Installer doesn't ring true to me, but I don't know what else does.  To me it's more like we are "building" the input/output tree from the cache (or vice versa), but don't know that Builder is any more intuitive.  Ultimately you will probably look at this code a lot more than I will, so I'm easy if you're confident in Installer.

pkg/proto/configuration/bb_worker/bb_worker.proto
EdSchouten: Let's just add that functionality when we get there. There's no harm in changing the schema, especially for features that were recently added.

jlaxson: Well, yes, but for the local code I have for followup that does

cmd/bb_worker/main.go
EdSchouten: FileInstaller, FilePlacer, FilePopulator, any will do. The point being: LinkHelper is inaccurate here, because it doesn't link files.

My suggestion was also to let "Cloning" and "Hardlinking" be prefixes instead of suffixes, as it's more consistent with the rest of the codebase.

pkg/cas/link_helper.go
```diff
@@ -0,0 +1,11 @@
+package cas
+
+import (
+	"github.com/buildbarn/bb-storage/pkg/filesystem"
+	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
+)
+
+// A LinkHelper implements a filesystem-specific method for copying files.
+type LinkHelper interface {
+	Link(directory filesystem.Directory, oldName path.Component, newDirectory filesystem.Directory, newName path.Component) error
```
EdSchouten, line 10: Nit: s/directory/oldDirectory/, for consistency with the other parameters? Also applies to link_helper_*.go.

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -110,6 +111,8 @@ message NativeBuildDirectoryConfiguration {
   // LEAST_RECENTLY_USED can, assuming the cache size is sufficient.
   buildbarn.configuration.eviction.CacheReplacementPolicy
       cache_replacement_policy = 5;
+
+  bool use_clonefile = 6;
```
EdSchouten, line 115: Could you please add some documentation on what this option does?

cmd/bb_worker/main.go
```diff
@@ -150,12 +150,19 @@ func main() {
 			if err != nil {
 				log.Fatal("Failed to create eviction set for cache directory: ", err)
 			}
-			fileFetcher = cas.NewHardlinkingFileFetcher(
+			var linkHelper cas.LinkHelper
+			if nativeConfiguration.UseClonefile {
```
EdSchouten, line 154: Be sure to abbreviate this to just:
```
linkHelper := cas.LinkHelperHardlink
if nativeConfiguration.UseCloneFile {
    linkHelper = cas.LinkHelperClonefile
}

File: pkg/cas/hardlinking_file_installer.go [hunk 10]
```diff
@@ -0,0 +1,15 @@
+package cas
+
+import (
+	"github.com/buildbarn/bb-storage/pkg/filesystem"
+	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
+)
+
+type hardlinkingFileInstaller struct{}
+
+// HardlinkingFileInstaller provides file copies through the use of link(2)
```
EdSchouten, line 10: Micronit: Could you add a period at the end of this sentence? :D

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -110,6 +111,12 @@ message NativeBuildDirectoryConfiguration {
   // LEAST_RECENTLY_USED can, assuming the cache size is sufficient.
   buildbarn.configuration.eviction.CacheReplacementPolicy
       cache_replacement_policy = 5;
+
+  // Use clonefile(2) to populate build action trees using APFS copy-on-
+  // write clones.  Only useful on Darwin when the build_directory_path
+  // is on an APFS volume.  When not set (default) uses standard hardlinks
+  // as supported by most all posix filesystems.
```
EdSchouten, line 118: s/posix/POSIX/

Maybe also add some explanation as why you'd want to set this option, namely to keep macOS security frameworks happy.

EdSchouten:
Is there going to be any further progress on this PR?

EdSchouten:
Closing due to inactivity.

PR #93, opened 2021-11-25
Make InMemoryBuildQueue timeouts configurable

Gormo:
Make scheduler timeouts configurable.


EdSchouten:
Hey there!

I sort of kept these non-configurable, because I wanted to figure out which options people actually want/need to have configurable. Which of these exact parameters do you want to tune for your setup? Would it be all right if we only exposed those?

EdSchouten:
> Thank you for the response! The one we would really like to configure is only **PlatformQueueWithNoWorkersTimeout** since that will trigger a hard fail on the bazel side if there's no queue available.

That seems fair. Do note that there's also a way to predeclare queues in config. That way they will never disappear.

> I'll update my patchset to only make that parameter configurable for now!

Note that Buildbarn binaries generally don't fill in default values for you. Considering that people are responsible for doing proper release engineering on their side when it comes to rolling out Buildbarn, there is absolutely no problem with having some breakage here and there. In other words, please make it so that we get a hard failure if no timestamp is provided, and make the existing value the recommended default.

Thanks!

Gormo:
> That seems fair. Do note that there's also a way to predeclare queues in config. That way they will never disappear.

Yes I've noticed that.. That does however force us to have the platform queues predeclared on the scheduler side which would be too much of a hassle in our case. We have a separate organization that maintains the scheduler/bb_storage side so that part should be as generic as possible.

> Note that Buildbarn binaries generally don't fill in default values for you

Thanks for that clarification! I was thinking a bit about what the strategy was since I didn't see any default-values elsewhere.

EdSchouten:
> > That seems fair. Do note that there's also a way to predeclare queues in config. That way they will never disappear.
> 
> Yes I've noticed that.. That does however force us to have the platform queues predeclared on the scheduler side which would be too much of a hassle in our case. We have a separate organization that maintains the scheduler/bb_storage side so that part should be as generic as possible.

Fair point!

> > Note that Buildbarn binaries generally don't fill in default values for you
> 
> Thanks for that clarification! I was thinking a bit about what the strategy was since I didn't see any default-values elsewhere.

Yep. The reason for it is that for a system like Buildbarn (that can either run on a Raspberry Pi or multiple racks of beefy servers), it's virtually impossible to come up with sane defaults most of the time. By making the fields required, people will actually give it all some thought.

EdSchouten:
Thanks!

Gormo:
@EdSchouten, I'm also thinking about adding a "StickyQueue" configuration that could be set on the worker-side and basically allows workers to create persistent queues on-the-fly when needed. Any opinions about that?

PR #95, opened 2022-01-18
Allow users of NewMountFromConfiguration to unmount cleanly

scele:
Allow users of NewMountFromConfiguration to cleanly unmount the FUSE
filesystem before process exit, by returning a Server interface to the
caller.


scele:
Accompanying bb-clientd PR: https://github.com/buildbarn/bb-clientd/pull/2

PR #99, opened 2022-03-13
PlatformOverridingSizeClassAnalyzer

jlaxson:
Some actions we would like to ensure receive consistent compute resources, for example benchmarks.  Adds a size class analyzer that can override the inferred size class based on a property specified in the platform properties.


pkg/scheduler/initialsizeclass/platform_overriding_size_class_analyzer.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (a *platformOverridingSizeClassAnalyzer) Analyze(ctx context.Context, digestFunction digest.Function, action *remoteexecution.Action) (Selector, error) {
+	timeout, err := a.actionTimeoutExtractor.ExtractTimeout(action)
+	if err != nil {
+		return nil, err
+	}
+
+	platform := action.Platform
+	for _, key := range platform.Properties {
```
EdSchouten, line 40: Should this also use platform.KeyExtractor, instead of trying to read the Platform message from Action directly? It could also be stored in the Command message.

pkg/scheduler/initialsizeclass/platform_overriding_size_class_analyzer.go
```diff
@@ -0,0 +1,91 @@
...
+	class, err := strconv.Atoi(s.sizeClass)
+	if err != nil {
+		return 0, s.timeout, learner, status.Errorf(codes.InvalidArgument, "Invalid size class '%s'", s.sizeClass)
+	}
+	for idx, val := range sizeClasses {
+		if int(val) == class {
+			return idx, s.timeout, learner, nil
+		}
+	}
+	return 0, s.timeout, learner, status.Errorf(codes.InvalidArgument, "Size class '%s' not available", s.sizeClass)
```
EdSchouten, line 75: I'm not sure that this is solid.

We currently only allow you to use feedback driven size class analysis by predeclaring size classes and providing a maximum size class. That way we can be certain that the system behaves deterministically, even if workers for larger size classes are (temporarily) offline. The client can be sure that if execution fails, it will always run on the largest worker.

The behaviour as implemented will cause hard execution failures, even if workers for a given size class are only temporarily offline. This may also be the case if the scheduler has just restarted and workers are still in the process of reconnecting.

pkg/proto/configuration/scheduler/scheduler.proto
```diff
@@ -294,3 +299,20 @@ message InitialSizeClassFeedbackDrivenAnalyzerConfiguration {
   // Recommended value: 32
   int32 history_size = 6;
 }
+
+message OverridingSizeClassAnalyzerConfiguration {
+  // The SizeClassAnalyzer to use when the specific platform key named
+  // below is not present.  If this is not provided, every action must
+  // specify the size class to use in its platform properties.
+  InitialSizeClassAnalyzerConfiguration base = 1;
```
EdSchouten, line 307: I think it's always the case you want to place PlatformOverridingSizeClassAnalyzer outside of FeedbackDrivenAnalyzer. Should we maybe get rid of the `oneof` above? As in, allow both fields to be set, but let 'overriding' wrap 'feedback_driven'.

EdSchouten:
Closing this due to inactivity. Feel free to reopen if needed. Thanks!

PR #101, opened 2022-03-16
Upgrade to Go 1.18

moroten:
This PR also removes some unused variables in tests, as that is required by the Go 1.18 compiler.


EdSchouten:
Thanks a lot!

PR #105, opened 2022-09-29
feat: add run command cleaner

joeljeske:
This adds an optional `run_command_cleaner` command that can be used to perform custom commands before/after an action to cleanup the action environment.


cmd/bb_runner/main.go
```diff
@@ -111,6 +111,15 @@ func main() {
 		}
 		cleaners = append(cleaners, cleaner.NewDirectoryCleaner(directory, d))
 	}
+
+	if configuration.RunCommandCleaner != nil {
```
EdSchouten, line 115: Just as a safety belt, could you change this to:
```go
if len(configuration.RunCommandCleaner) > 0 {
```
I'm pretty sure that `proto.Unmarshal()` leaves empty lists as nil, but we can't be too sure about that. A non-nil empty list would cause us to index the list out of bounds.

pkg/cleaner/exec_process_cleaner.go
```diff
@@ -0,0 +1,14 @@
+package cleaner
+
+import (
+	"context"
+	"os/exec"
+)
+
+// NewExecProcessCleaner creates a new Cleaner that exec's an command on the
+// system and expects a 0 exit code.
+func NewExecProcessCleaner(command string, args []string) Cleaner {
```
EdSchouten, line 10: What are your thoughts on calling this NewCommandRunningCleaner()?

pkg/cleaner/exec_process_cleaner.go
```diff
@@ -0,0 +1,14 @@
...
+	return func(ctx context.Context) error {
+		return exec.CommandContext(ctx, command, args...).Run()
```
EdSchouten, line 12: I was going to ask you: Why can't we let NewExecProcessCleaner() take an *exec.Cmd, so that we can clone it as part of this function? That way consumers of this Cleaner may also override the working directory, environment variables, SysProcAttr, etc. It turns out that exec.Cmd is a bit of a mess:

- It contains both inputs (environment variables, arguments) and outputs (process state, LookPath errors),
- There is no way to override the `ctx` after it's been created.

So yeah. Let's stick to this approach. :-/

One recommendation I have, though. Could you change this to something like this?
```go
if err := exec.CommandContext(ctx, command, args...).Run(); err != nil {
     return util.StatusWrapWithCode(err, codes.Internal, "Failed to run cleaning command")
}
return nil
```
Otherwise it may not be obvious from the logs that any failures are caused by this feature.

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -76,4 +76,13 @@ message ApplicationConfiguration {
...
   // only up to a single command concurrently.
   repeated string symlink_temporary_directories = 12;
+
+  // Run a command with arguments to clean the system for actions
+  // This commmand runs at least once before an action, and once immediately
+  // after an action.
+  //
+  // Caution; since this option executes a command in a separate process
+  // there is an inherent performance penalty. If possible use the built-in
+  // cleaning options that are handled in-process for optimal performance
```
EdSchouten, line 86: (Nit: Missing full stop at the end of the sentence.)

Thanks for adding this notice!

pkg/cleaner/exec_process_cleaner.go
```diff
@@ -0,0 +1,14 @@
+package cleaner
+
+import (
+	"context"
+	"os/exec"
+)
+
+// NewExecProcessCleaner creates a new Cleaner that exec's an command on the
```
EdSchouten, line 8:
```go
// NewExecProcessCleaner creates a new Cleaner that executes a command on the
// system and expects it to succeed.
```
?

EdSchouten (commented):
This is already looking pretty good. Thanks! Please take a look at some of the comments I left behind.

pkg/cleaner/exec_process_cleaner.go
```diff
@@ -0,0 +1,14 @@
+package cleaner
```
EdSchouten, line 1: Could we also have some very basic unit testing coverage for this code? It can be pretty simple: Just use it to run `true` and `false`.

pkg/cleaner/exec_process_cleaner.go
joeljeske: Yes, thats a great callout! I had not considered this.

pkg/cleaner/exec_process_cleaner.go
joeljeske: Done!

pkg/cleaner/command_running_cleaner_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func TestCommandRunningCleaner(t *testing.T) {
+	_, ctx := gomock.WithContext(context.Background(), t)
+
+	t.Run("Failure", func(t *testing.T) {
+		testutil.RequireEqualStatus(
+			t,
+			status.Error(codes.Internal, "Failed to run cleaning command: exit status 1"),
+			cleaner.NewCommandRunningCleaner("false", []string{})(ctx))
```
EdSchouten, line 23: Nit: You can just pass in `nil` instead of `[]string{}`. The same below.

cmd/bb_runner/main.go
```diff
@@ -111,6 +111,15 @@ func main() {
 		}
 		cleaners = append(cleaners, cleaner.NewDirectoryCleaner(directory, d))
 	}
+
+	if len(configuration.RunCommandCleaner) > 0 {
+		cleaners = append(
+			cleaners,
+			cleaner.NewExecProcessCleaner(
```
EdSchouten, line 118: NewCommandRunningCleaner

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -76,4 +76,13 @@ message ApplicationConfiguration {
   // It is only safe to enable this option when this runner executes
   // only up to a single command concurrently.
   repeated string symlink_temporary_directories = 12;
+
+  // Run a command with arguments to clean the system for actions
```
EdSchouten, line 80: Could you clarify this sentence a bit? It looks a bit incomplete.

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -76,4 +76,13 @@ message ApplicationConfiguration {
...
+  // This commmand runs at least once before an action, and once immediately
+  // after an action.
```
EdSchouten, line 82: I think it's a bit more correct to say:

> This command runs when the runner transitions from being idle to running at least one action, and vice versa. It is also called periodically whenever the runner is idle.

PR #107, opened 2022-12-20
Upgrade to Bazel 6.0 and all the Go dependencies

moroten:


go.mod
```diff
@@ -11,51 +11,50 @@ replace github.com/gordonklaus/ineffassign => github.com/gordonklaus/ineffassign
 replace mvdan.cc/gofumpt => mvdan.cc/gofumpt v0.2.1
```
moroten, line 11: Should be v0.3.0

moroten:
@EdSchouten are you happy with this change?

EdSchouten:
Hey! Yes, looks good. But let me first fix the CI failure in bb-storage before merging this.

pkg/blobstore/blob_access_mutable_proto_store.go
```diff
@@ -75,7 +75,8 @@ type blobAccessMutableProtoStore[T any, TProto interface {
 func NewBlobAccessMutableProtoStore[T any, TProto interface {
 	*T
 	proto.Message
-}](initialSizeClassCache blobstore.BlobAccess, maximumMessageSizeBytes int) MutableProtoStore[TProto] {
+}](initialSizeClassCache blobstore.BlobAccess, maximumMessageSizeBytes int,
```
EdSchouten, line 78: What's going on here, by the way? Looks like an odd way to format this code.

EdSchouten: Maybe we should leave gofumpt pinned at 0.2.1?

EdSchouten (commented):
bb-storage now builds fine, with a minor change to the GitHub workflow definitions. Could you update go.mod to use the latest bb-storage?

pkg/blobstore/blob_access_mutable_proto_store.go
moroten: bb-storage and the others have 0.3.0. Seems like I have to run gofumpt twice to get it stable, splitting the line the second time.

moroten:
Should we add a `.bazelversion` file to the Buildbarn repositories?

EdSchouten:
I don't know. That's a Bazelisk specific thing, right? I don't use Bazelisk myself, but if it helps folks using it, sure. :-)

moroten:
`.bazelversion` is actually parsed by the "installed" Bazel to verify it is the correct version. If you use the binary download directly no such check is done. One can also add [versions.check](https://github.com/bazelbuild/bazel-skylib/blob/1.3.0/docs/versions_doc.md#versionscheck) to the `WORKSPACE` to really force a single version, but I think that is too picky for a project like Buildbarn. Let's think about pros and cons for the next time.

moroten:
I'm happy to merge this PR in its current state.

pkg/blobstore/blob_access_mutable_proto_store.go
EdSchouten: I'd rather stick to 0.2.1 in this repo if that prevents the unnecessary formatting.

moroten: OK, fixing...

moroten: Went back to gofumpt 0.2.1 and added a comment. Also checked 0.4.0 but it formats the same way as 0.3.0.

PR #111, opened 2023-03-22
Allow passing arbitrary options to fusermount

jmmv:
Extend the FUSE mount options to support passing a collection of textual options that are directly passed to fusermount.

Fixes https://github.com/buildbarn/bb-clientd/issues/6.


pkg/proto/configuration/filesystem/virtual/virtual.proto
```diff
@@ -133,6 +133,9 @@ message FUSEMountConfiguration {
   //
   // Recommended value: unset
   map<string, string> linux_backing_dev_info_tunables = 9;
+
+  // List of options to pass directly to fusermount.
+  repeated string options = 10;
```
EdSchouten, line 138: Could we describe what the semantics of this option are if direct_mount is used?

EdSchouten:
Closing this due to inactivity.

PR #116, opened 2023-09-06
Mount special filesystem in chroot runners

stagnation:
This mounts '/proc' and '/sys' in the input root of runners if 'chroot' is enabled. It is done with "at"-syscall semantics in the 'Directory', to avoid side effects. A program-scope mutex is required for unmounting because there is no easy-to-use "unmountat".
https://github.com/buildbarn/bb-remote-execution/issues/115


pkg/runner/local_runner.go
```diff
@@ -136,6 +157,44 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
...
 	cmd.Stdout = stdout
 
+	// Mount special filesystems.
+	requiresSpecialFilesystems := cmd.SysProcAttr != nil && cmd.SysProcAttr.Chroot != ""
+	var MountRootDirectory filesystem.Directory
+
+	if requiresSpecialFilesystems {
+		var dir filesystem.Directory
+		dir = r.buildDirectory
+		// TODO(nils): How to best open the input root `Directory`
```
stagnation, line 167: This seems to be a misunderstanding on my part.
I want (to hold on to) a `Directory` object for the input root.
It can be reopened, but do not want to call this opening code twice, I don't think.

pkg/runner/local_runner.go
```diff
@@ -136,6 +157,44 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
...
+		// It must be a `localDirectory`, but outer directories can be `lazyDirectory`.
```
stagnation, line 168: This strings-split is a cop out for my lack of understanding of the builder and scope walker.
I want to replace this block.

:-1:

pkg/runner/local_runner.go
```diff
@@ -166,6 +225,14 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
 		}
 	}
 
+	if requiresSpecialFilesystems {
```
stagnation, line 228: This is not defered, as I saw some unmounting errors during development that would not surface well.
Do you want me to create a defer-lambda that modifes the `err` on the way out to set an error
or is this an okay clean-up section?

stagnation (commented):
:-1:

To my reading the mount calls require 'CAP_SYS_ADMIN', there is no 'CAP_MOUNT', so this may have consequences for deployments. But then again, to use `chroot` where the tools crash is also ill-advised.

pkg/runner/local_runner.go
```diff
@@ -136,6 +157,44 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
...
+		// It must be a `localDirectory`, but outer directories can be `lazyDirectory`.
+		for _, segment := range strings.Split(request.InputRootDirectory, "/") {
+			component := path.MustNewComponent(segment)
+			dircloser, err := dir.EnterDirectory(component)
+			// We want to keep the input root open so we can use its file descriptor.
+			if segment != "root" {
+				defer dircloser.Close()
+			}
+
+			dir = dircloser.(filesystem.Directory)
```
EdSchouten, line 177: Doesn't DirectoryCloser implement Directory?

pkg/runner/local_runner.go
EdSchouten: Yeah, having it as a `defer func() {}()` makes sense.

pkg/runner/local_runner.go
EdSchouten: Yeah, you likely also don't want it, because the directory can be moved around to other places, right?

pkg/runner/local_runner.go
EdSchouten: No worries. I can also help out with this part, after we get the bb-storage part sorted out.

EdSchouten (commented):
Thanks for working on this, Nils!

pkg/runner/local_runner.go
```diff
@@ -136,6 +154,44 @@ func (r *localRunner) Run(ctx context.Context, request *runner.RunRequest) (*run
...
+			if err != nil {
+				return nil, fmt.Errorf("Could not enter directory component '%s' in '%#v'", segment, inputRootDirectory)
+			}
+		}
+		MountRootDirectory = dir
+
+		for _, mount := range mounts {
+			err := MountRootDirectory.Mkdir(mount.mountpoint, 0o555)
+			if err != nil {
+				if err != unix.EEXIST {
```
EdSchouten, line 184: os.IsExist() ?

pkg/runner/local_runner.go
```diff
@@ -3,20 +3,38 @@ package runner
...
+// to mount in the input root when using 'chroot'.
+var (
+	procMountpointComponent = path.MustNewComponent("proc")
+	sysMountpointComponent  = path.MustNewComponent("sys")
+	mounts                  = []struct {
+		mountpoint path.Component
+		fstype     string
+		source     string
+	}{
+		{procMountpointComponent, "proc", "/proc"},
```
EdSchouten, line 33: Should this be specified in config?
```proto
repeated MountPointConfiguration mount_points = 123;
```

In main.go:
```go
for _, mountPoint := range configuration.MountPoints {
     runner = NewMountingRunner(runner, mountPoint)
}
```
?

EdSchouten (commented):
General remark: Could this be implemented as a decorator for Runner?

```go
type mountingRunner struct {
     base Runner
}
```

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,102 @@
...
+	base        runner_pb.RunnerServer
+	buildDirectory filesystem.Directory
+	mount *bb_runner.MountInfo
+}
+
+// NewMountingRunner is a decorator for Runner
+// that mounts `mount` before running a build action.
+//
+// This decorator can be used for chroot runners
+// that must mount special filesystems into the input root.
```
stagnation, line 28: We could also compare it to the clean runner that has generalized pre- and post-command execution.

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+
+	if err := root.Mount(mountpoint, r.mount.Source, r.mount.FilesystemType); err != nil {
+		return nil, util.StatusWrapf(err, "Failed to mount '%#v' in the input root", r.mount)
+	}
+	defer func() {
+		// TODO: is `root` open here or do we need to open it again?
+		if err2 := root.Unmount(mountpoint); err2 != nil {
+			err = util.StatusFromMultiple([]error{
```
stagnation, line 85: is this right? if there was not preceding error we should not have multiple. Maybe the function is helpful and just returns one here, or we should add a check.

pkg/runner/local_runner.go
stagnation: Yepp!

pkg/runner/mounting_runner.go
stagnation: added a check

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -131,4 +131,23 @@ message ApplicationConfiguration {
...
+  // some tools require access to special filesystems
+  // that are created when the operating system boots.
+  // An input root with a full userland implementation may need these.
+  // Typical choices are:
+  //    {"proc", "/proc", "proc"}
+  //    {"sys", "/sys", "sysfs"}
+  repeated mount_info input_root_mounts = 15;
+}
+
+message mount_info {
```
EdSchouten, line 146: Could you please use PascalNaming here? InputMountOptions or something?

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -131,4 +131,23 @@ message ApplicationConfiguration {
   // https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/exec/local/XcodeLocalEnvProvider.java
   // https://www.smileykeith.com/2021/03/08/locking-xcode-in-bazel/
   map<string, string> apple_xcode_developer_directories = 14;
+
+  // Mount special filesystems in the input root.
+  // This is useful when running with `chroot`.
```
EdSchouten, line 136: > This is useful in combination with `chroot_into_input_root`.

Could you please wrap this documentation at ~70-80 columns?

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -131,4 +131,23 @@ message ApplicationConfiguration {
...
+  // some tools require access to special filesystems
+  // that are created when the operating system boots.
+  // An input root with a full userland implementation may need these.
+  // Typical choices are:
+  //    {"proc", "/proc", "proc"}
```
EdSchouten, line 141: Please provide full JSON excerpts here, including the keys of the fields.

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	for i, segment := range parts {
+		var err error
+		component := path.MustNewComponent(segment)
+		dir, err = iterator.EnterDirectory(component)
+		if i < len(parts) - 1 {
+			defer dir.Close()
+		}
+
+		iterator = dir.(filesystem.Directory)
+		if err != nil {
```
EdSchouten, line 57: This error handling is in the wrong place.

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func open(buildDirectory filesystem.Directory, inputRootDirectory string) (filesystem.DirectoryCloser, error) {
+	// TODO(nils): How to best open the input root `Directory`
+	//             It must be a `localDirectory`, but outer directories can be `lazyDirectory`.
+	// NB: We iterate over Directory, as the initial buildDirectory is not a DirectoryCloser.
+	iterator := buildDirectory
+	var dir filesystem.DirectoryCloser
+
+	parts := strings.Split(inputRootDirectory, "/")
```
EdSchouten, line 44: Please use `path.Resolve()` here.

```go
if err := path.Resolve(thePath, path. NewRelativeScopeWalker(componentWalker)); err != nil {
     ...
}
```

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	if len(parts) == 0 {
```
EdSchouten, line 45: The input root directory is in theory permitted to be empty. There is nothing in the runner protocol that forbids it. It may also be ".".

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, status.Error(codes.FailedPrecondition, "Unexpected empty inputRootDirectory")
+	}
+	for i, segment := range parts {
+		var err error
+		component := path.MustNewComponent(segment)
```
EdSchouten, line 50: This should use a regular `NewComponent()`. You can't assume the provided path is well-formed.

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (r *mountingRunner) Run(ctx context.Context, request *runner_pb.RunRequest) (response *runner_pb.RunResponse, err error) {
+	root, err := open(r.buildDirectory, request.InputRootDirectory)
+	if err != nil {
+		return nil, status.Errorf(codes.FailedPrecondition, "Could not enter input root directory '%s'", request.InputRootDirectory)
+	}
+	defer root.Close()
+
+	mountpoint := path.MustNewComponent(r.mount.Mountpoint)
```
EdSchouten, line 71: What about mounts in some nested location?

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	err = root.Mkdir(mountpoint, 0o555)
```
EdSchouten, line 72: Should we be creating the directories ourselves? Maybe it makes more sense to only establish these mounts if they actually exist in the input root?

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+		if ! os.IsExist(err) {
+			return nil, util.StatusWrapf(err, "Failed to create mount point: '%#v' in the input root", r.mount.Mountpoint)
+		}
+	}
+
+	if err := root.Mount(mountpoint, r.mount.Source, r.mount.FilesystemType); err != nil {
+		return nil, util.StatusWrapf(err, "Failed to mount '%#v' in the input root", r.mount)
+	}
+	defer func() {
+		if err2 := root.Unmount(mountpoint); err2 != nil {
```
EdSchouten, line 83: With only the `Run()` call in between, I would not recommend using `defer` here. Just place this code right after calling `Run()`.

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.FailedPrecondition, "Could not enter input root directory '%s'", request.InputRootDirectory)
+	}
+	defer root.Close()
+
+	mountpoint := path.MustNewComponent(r.mount.Mountpoint)
+	err = root.Mkdir(mountpoint, 0o555)
+	if err != nil {
+		if ! os.IsExist(err) {
+			return nil, util.StatusWrapf(err, "Failed to create mount point: '%#v' in the input root", r.mount.Mountpoint)
```
EdSchouten, line 75: There is no need to put quotes around strings if you're using `%#v`. It already does that for you.

pkg/runner/mounting_runner.go
stagnation: Good point, with this as an advanced feature, we can probably require very strict compliance from clients that wants to use it. I can require that in the proto specification.

pkg/runner/mounting_runner.go
stagnation: Good point! They could be in any directory hierarchy, the code does not support that as it is written yet.

Aside, nested mounts: the mounts themselves can also be nested, If one wants to use the original filesystem as a mountpoint the innermost path must be given first.

```
mkdir -p outer/inner/x
mount -t sysfs /sys outer/inner
mount -t sysfs /sys outer
mount -t sysfs /sys outer/inner/x  # x is not inside the mounted sysfs
mount -t sysfs /sys outer/inner/block # can be mounted ontop of the other sysfs.
```

cmd/bb_runner/main.go
```diff
@@ -68,6 +68,14 @@ func main() {
 			buildDirectoryPath,
 			commandCreator,
 			configuration.SetTmpdirEnvironmentVariable)
+		for _, mountinfo := range configuration.InputRootMounts {
```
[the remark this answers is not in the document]
EdSchouten: I actually prefer the way the code is written right now. Nothing wrong with looping over an empty list.

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -131,4 +131,37 @@ message ApplicationConfiguration {
   // https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/exec/local/XcodeLocalEnvProvider.java
   // https://www.smileykeith.com/2021/03/08/locking-xcode-in-bazel/
   map<string, string> apple_xcode_developer_directories = 14;
+
+  // Mount special filesystems in the input root. This is useful when
+  // running with `chroot_into_input_root`. some tools require access to
```
EdSchouten, line 136: s/some/Some/

pkg/proto/configuration/bb_runner/bb_runner.proto
```diff
@@ -131,4 +131,37 @@ message ApplicationConfiguration {
...
+  //       filesystemType: 'sysfs',
+  //     },
+  //   ],
+  repeated InputMountOptions input_root_mounts = 15;
+}
+
+message InputMountOptions {
+  // Mount a filesystem in the input root, a relative path.
+  string mountpoint = 1;
+  // Source filesystem from the runner's operating system, absolute path.
```
EdSchouten, line 163: I wouldn't document this as being an absolute path. What this string needs to be is actually specific on the type of file system used. In the case of NFS it would be of the form 192.168.1.1:/foo.

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (r *mountingRunner) Run(ctx context.Context, request *runner_pb.RunRequest) (response *runner_pb.RunResponse, err error) {
+	pathResolver := buildDirectoryPathResolver{
+		stack: util.NewNonEmptyStack(filesystem.NopDirectoryCloser(r.buildDirectory)),
+	}
+	scopeWalker := path.NewRelativeScopeWalker(&pathResolver)
+	defer pathResolver.closeAll()
+
+	fullpath := filepath.Join(request.InputRootDirectory, r.mount.Mountpoint)
```
EdSchouten, line 43: My suggestion would be to not use `filepath.Join()` here. It assumes that the native path separator is used, even though `request.InputRootDirectory` always uses forward slashes, even on Windows. There is also no need to do that if you just call `path.Resolve()` multiple times:
```go
if err := path.Resolve(request.InputRootDirectory, path.NewRelativeScopeWalker(&pathResolver)); err != nil {
    ...
}
if err := path.Resolve(r.mount.Mountpoint, path.NewRelativeScopeWalker(&pathResolver)); err != nil {
    ...
}
```

I would also strongly advise against using the same PathResolver for both processing the input root directory and the mount point. Notice how the path resolver contains a stack. That stack is there to process `".."` entries. Because the stack isn't cleared/truncated, a mount point containing `".."` components might leave the input root directory.

This is not a big issue right now, because the mount point is fully specified in config. But if someone ever comes along and wants the mount point path resolution to respect symlinks, it can become dangerous fairly quickly.

pkg/runner/mounting_runner.go
```diff
@@ -0,0 +1,75 @@
...
+		base:           base,
+	}
+}
+
+func (r *mountingRunner) Run(ctx context.Context, request *runner_pb.RunRequest) (response *runner_pb.RunResponse, err error) {
+	pathResolver := buildDirectoryPathResolver{
+		stack: util.NewNonEmptyStack(filesystem.NopDirectoryCloser(r.buildDirectory)),
+	}
+	scopeWalker := path.NewRelativeScopeWalker(&pathResolver)
+	defer pathResolver.closeAll()
```
EdSchouten, line 41: This might keep more file descriptors open than strictly necessary. Would there be a way for us to close intermediate directories while `Run()` is called?

EdSchouten (commented):
Thanks for working on this!

pkg/proto/configuration/bb_runner/bb_runner.proto
stagnation: :eyes: or my favorite metacharacter: s/some/\u&/

stagnation:
I think this is a flaky storage error in the check, but other storage error that occurred around the same time had more explicit error messages.
```
ERROR: /home/runner/work/bb-remote-execution/bb-remote-execution/cmd/bb_scheduler/BUILD.bazel:93:9: ImageLayer cmd/bb_scheduler/bb_scheduler_container-layer.tar failed: (Exit 1): build_tar failed: error executing command (from target //cmd/bb_scheduler:bb_scheduler_container) bazel-out/k8-opt-exec-2B5CBBC6-ST-625e526ca8a8/bin/external/io_bazel_rules_docker/container/build_tar ... (remaining 5 arguments skipped)
```

stagnation:
With https://github.com/buildbarn/bb-remote-execution/pull/225 we'll have more reason to use this :)

moroten:
Instead of `unmountat`, would a mount namespace help?

PR #119, opened 2023-11-18
Drop CAP_SYS_ADMIN after FUSE initialization

mortenmj:
Mounting a FUSE filesystem on Linux requires the CAP_SYS_ADMIN capability, which gives very broad access (see https://man7.org/linux/man-pages/man7/capabilities.7.html). Dropping CAP_SYS_ADMIN when we no longer need it should reduce the risk involved with having to grant this capability in the first place.

### DIFF
--- a/pkg/filesystem/virtual/configuration/fuse_mount_enabled.go
+++ b/pkg/filesystem/virtual/configuration/fuse_mount_enabled.go
@@ -94,5 +94,11 @@ func (m *fuseMount) Expose(terminationGroup program.Group, rootDirectory virtual
 	); err != nil {
 		return util.StatusWrap(err, "Failed to set Linux Backing Device Info tunables")
 	}
+
+	// Access to call mount() is no longer needed now that FUSE has been initialized.
+	if err := dropMountCapabilities(); err != nil {
+		return util.StatusWrap(err, "Unable to drop mount capability. You may be missing CAP_SETPCAP")
+	}
+
 	return nil
 }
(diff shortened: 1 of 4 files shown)


pkg/filesystem/virtual/configuration/fuse_mount_enabled.go
```diff
@@ -94,5 +94,11 @@ func (m *fuseMount) Expose(terminationGroup program.Group, rootDirectory virtual
 	); err != nil {
 		return util.StatusWrap(err, "Failed to set Linux Backing Device Info tunables")
 	}
+
+	// Access to call mount() is no longer needed now that FUSE has been initialized.
```
EdSchouten, line 98: The downside of this approach is that it doesn't compose well. For example, you can set up multiple build directories. In fact, for FUSE based workers it's actually not uncommon to do that. That way each worker thread is only capable of seeing its own input files. Also, if you refactor your worker config to not use FUSE at all, you'll never end up calling this.

I think we need to introduce some global API (pkg/global?) that allows us to specify 'leases' on capabilities, and have individual subsystems drop these leases. If the number of leases drops to zero, we can actually call `prctl()` to drop them.

stagnation: I am fairly certain we need it for chroot runners too. It is high time I dust that off. But I really appreciate the pledge mentality. Would be very interesting to see how it can be done within Buildbarn.

PR #123, opened 2024-01-14
Copyfile hardlink fallback

jacobshirley:
I've found that on MacOS, hardlinking PkgInfo and lproj files does not work. This fix falls back to copy files to the cache rather than hardlinking


EdSchouten:
Why does hardlinking not work? What is the exact error message you are getting?

Furthermore, on macOS I strongly recommend using bb_worker's integrated NFSv4 file system. If you don't use it, there is also a fair chance you'll run into dyld related issues.

PR #125, opened 2024-02-07
Tidy up

stagnation:
Some small changes to I made during the work to update windows workers.
Keep them as a separate PR to merge them sooner, and reduce the scope of the meaty review of the PR.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -175,9 +175,9 @@ http_archive(
 
 http_archive(
     name = "aspect_rules_js",
-    sha256 = "00e7b97b696af63812df0ca9e9dbd18579f3edd3ab9a56f227238b8405e4051c",
-    strip_prefix = "rules_js-1.23.0",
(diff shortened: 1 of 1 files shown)


cmd/bb_worker/main.go
```diff
@@ -254,10 +254,10 @@ func main() {
 				// used. This increases cache hit rate.
 				cacheDirectory, err := filesystem.NewLocalDirectory(nativeConfiguration.CacheDirectoryPath)
 				if err != nil {
-					return util.StatusWrap(err, "Failed to open cache directory")
+					return util.StatusWrapf(err, "Failed to open cache directory: %v", nativeConfiguration.CacheDirectoryPath)
```
EdSchouten, line 257:
```go
return util.StatusWrapf(err, "Failed to open cache directory %#v", nativeConfiguration.CacheDirectoryPath)
```
Drop the `:` and add a `#`. The `#` causes the string to be quoted. Same in other places in this PR.

PR #126, opened 2024-02-08
Feature/freshen up windows workers

stagnation:
.

### DIFF
--- a/pkg/builder/local_build_executor.go
+++ b/pkg/builder/local_build_executor.go
@@ -160,7 +160,14 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 			util.StatusWrap(err, "Failed to acquire build environment"))
 		return response
 	}
-	defer buildDirectory.Close()
+	defer func() {
+		err := buildDirectory.Close()
+		if err != nil {
+			attachErrorToExecuteResponse(
+				response,
+				util.StatusWrap(err, "Failed to close build directory"))
+		}
+	}()
 
 	// Install hooks on build directory to capture file creation and
 	// I/O error events.
--- a/pkg/runner/local_runner_windows.go
+++ b/pkg/runner/local_runner_windows.go
@@ -4,14 +4,11 @@
 package runner
 
 import (
-	"context"
 	"os"
 	"os/exec"
 	"syscall"
 
 	"github.com/buildbarn/bb-remote-execution/pkg/proto/resourceusage"
-	"github.com/buildbarn/bb-storage/pkg/filesystem/path"
-	"github.com/buildbarn/bb-storage/pkg/util"
 
 	"golang.org/x/sys/windows"
 	"google.golang.org/grpc/codes"
@@ -19,34 +16,10 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
-// NewPlainCommandCreator returns a CommandCreator for cases where we don't
-// need to chroot into the input root directory.
-func NewPlainCommandCreator(sysProcAttr *syscall.SysProcAttr) CommandCreator {
-	return func(ctx context.Context, arguments []string, inputRootDirectory *path.Builder, workingDirectoryParser path.Parser, pathVariable string) (*exec.Cmd, error) {
-		// TODO: This may not work correctly if the action sets
-		// the PATH environment variable explicitly.
-		cmd := exec.CommandContext(ctx, arguments[0], arguments[1:]...)
-		cmd.SysProcAttr = sysProcAttr
-
-		// Set the working relative to be relative to the input
-		// root directory.
-		workingDirectory, scopeWalker := inputRootDirectory.Join(path.VoidScopeWalker)
-		if err := path.Resolve(workingDirectoryParser, scopeWalker); err != nil {
-			return nil, util.StatusWrap(err, "Failed to resolve working directory")
-		}
-		workingDirectoryStr, err := path.GetLocalString(workingDirectory)
-		if err != nil {
-			return nil, util.StatusWrap(err, "Failed to create local representation of working directory")
-		}
-		cmd.Dir = workingDirectoryStr
-		return cmd, nil
-	}
-}
-
 // NewChrootedCommandCreator gives an error on Windows, as chroot is not
 // supported on the platform.
 func NewChrootedCommandCreator(sysProcAttr *syscall.SysProcAttr) (CommandCreator, error) {
-	return nil, status.Error(codes.InvalidArgument, "Chroot not supported on Windows")
+	return nil, status.Error(codes.InvalidArgument, "Chroot is not supported on Windows")
 }
 
 var temporaryDirectoryEnvironmentVariablePrefixes = [...]string{"TMP=", "TEMP="}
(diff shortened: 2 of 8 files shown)


pkg/builder/local_build_executor.go
```diff
@@ -127,9 +128,9 @@ func (be *localBuildExecutor) CheckReadiness(ctx context.Context) error {
 	return err
 }
 
-func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesystem.FilePool, monitor access.UnreadDirectoryMonitor, digestFunction digest.Function, request *remoteworker.DesiredState_Executing, executionStateUpdates chan<- *remoteworker.CurrentState_Executing) *remoteexecution.ExecuteResponse {
+func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesystem.FilePool, monitor access.UnreadDirectoryMonitor, digestFunction digest.Function, request *remoteworker.DesiredState_Executing, executionStateUpdates chan<- *remoteworker.CurrentState_Executing) (response *remoteexecution.ExecuteResponse) {
 	// Timeout handling.
-	response := NewDefaultExecuteResponse(request)
+	response = NewDefaultExecuteResponse(request)
```
EdSchouten, line 133: Why is this change needed? If it's not needed, please revert.

pkg/builder/local_build_executor.go
```diff
@@ -160,7 +161,12 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 			util.StatusWrap(err, "Failed to acquire build environment"))
 		return response
 	}
-	defer buildDirectory.Close()
+	defer func() {
+		err := buildDirectory.Close()
+		if err != nil {
+			log.Printf("Failed to clean up build environment: %v\n", err)
+		}
+	}()
```
EdSchouten, line 169: Can't we simply do:
```go
defer func() {
    if err := buildDirectory.Close(); err != nil {
        attachErrorToExecuteResponse(response, util.StatusWrap(err, "Failed to close build directory"))
    }
}()
```
?

pkg/runner/local_runner_windows.go
```diff
@@ -25,6 +26,24 @@ func NewPlainCommandCreator(sysProcAttr *syscall.SysProcAttr) CommandCreator {
 	return func(ctx context.Context, arguments []string, inputRootDirectory *path.Builder, workingDirectoryParser path.Parser, pathVariable string) (*exec.Cmd, error) {
```
EdSchouten: If we can't get rid of this CommandCreator implementation, can we at least ensure it's as similar to the one we use on UNIX as possible, in terms of comments, ordering of statements, etc. etc. etc.

pkg/runner/local_runner_windows.go
```diff
@@ -25,6 +26,24 @@ func NewPlainCommandCreator(sysProcAttr *syscall.SysProcAttr) CommandCreator {
...
+		pathBuilder, scopeWalker := path.EmptyBuilder.Join(path.VoidScopeWalker)
+		if err := path.Resolve(path.NewLocalParser(argv0), scopeWalker); err != nil {
+			return nil, err
+		}
+		argv0, err := pathBuilder.GetWindowsString()
+		if err != nil {
+			return nil, err
+		}
+
+		arguments[0] = "\\nonexistent-place-back-later"
```
EdSchouten: This is mutating one of the input arguments of this function. Is this safe/desirable? Considering that we call `exec.CommandContext()` with `arguments[0]` and `arguments[1:]` separately, why not just pass in the string there?

pkg/runner/local_runner_windows.go
```diff
@@ -54,6 +86,7 @@ var temporaryDirectoryEnvironmentVariablePrefixes = [...]string{"TMP=", "TEMP="}
 var invalidArgumentErrs = [...]error{exec.ErrNotFound, os.ErrPermission, os.ErrNotExist, windows.ERROR_BAD_EXE_FORMAT}
 
 func getPOSIXResourceUsage(cmd *exec.Cmd) *resourceusage.POSIXResourceUsage {
+	// TODO: These do not work
```
EdSchouten, line 30: Please end all sentences with a full stop.

pkg/runner/local_runner_windows.go
stagnation: Yes. With the shared command creator that is much better. This was an artifact from also setting the argv0 on Windows to match cmd.Path. But It could have been implemented better. Now though we do not choose to do so.

pkg/runner/local_runner_windows.go
stagnation: Yepp!

pkg/builder/local_build_executor.go
stagnation: Yes. I have a commit to do just that, and am looking into a better way to avoid Bazel retrying and printing follow up errors to the user.

This is the commit message if you find it useful, or I can prune it if it is too much information.

```
    Send failure to clean up build directory to the client

    The client may choose to retry the action and then receive a different
    follow-up error.

        Action: http://.../remote-execution/blobs/sha256/action/6068c5370fdcca9dfb70936f5ea481cf53092c5c3ea328e351a69641d5c7fbfa-169/
        ExecuteResponse: {...,
            "status":{
                "code":13,
                "message":"Failed to close build directory: Failed to remove build directory \"6068c5370fdcca9d\": Failed to remove \"root/external/com_google_absl/absl/base/internal\": directory not empty"
            }}

        Action: http://.../remote-execution/blobs/sha256/action/6068c5370fdcca9dfb70936f5ea481cf53092c5c3ea328e351a69641d5c7fbfa-169/
        ExecuteResponse: {...,
            "status":{
                "code":13,
                "message":"Failed to acquire build environment: Failed to create build directory \"6068c5370fdcca9d\": Cannot create a file when that file already exists."
        }}
```

pkg/builder/local_build_executor.go
stagnation: No it is not needed :+1:. I assumed when first trying to modify the response in the defer closure that we needed to have the name captured as a named output. Then forgot it when I rewrote the code to just log, and now that we send the error again I find that this change is not needed.

pkg/builder/local_build_executor.go
```diff
@@ -2,6 +2,7 @@ package builder
 
 import (
 	"context"
+	"log"
```
EdSchouten, line 5: I think this should be removed.

pkg/builder/local_build_executor.go
```diff
@@ -160,7 +161,14 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool re_filesyste
 			util.StatusWrap(err, "Failed to acquire build environment"))
 		return response
 	}
-	defer buildDirectory.Close()
+	defer func() {
+		err := buildDirectory.Close()
```
EdSchouten, line 165:
```
if err := buildDirectory.Close(); err != nil {
```

pkg/builder/local_build_executor.go
stagnation: Concurrency is also an important note here if we want to accurately describe the current state of affairs, or we can punt that and look at the required improvements instead.

PR #131, opened 2024-03-26
upgrade bb-storage and Bazel

alexeagle:
Pre-factor for working on #129


alexeagle:
oh I see you already did this @EdSchouten

PR #138, opened 2024-06-12
Feature/use path.parser for new local directory

stagnation:
https://github.com/buildbarn/bb-storage/pull/206


stagnation:
https://github.com/buildbarn/bb-remote-execution/commit/2bcc408397c4653598e2714a0206f909948b9b48

PR #151, opened 2024-10-09
Expose platform properties in scheduler metrics

alexnovak:
Currently, prometheus metrics on the scheduler report all platform properties as a json-serialized string under the "properties" label. This is useful for bb-autoscaler, which knows how to parse this value, but is very challenging to use with tooling which might already be available for autoscaling in your setup, like [keda](https://keda.sh/). It also makes it challenging to dashboard or analyze queue metrics.

This change modifies the metric reporting behavior in the scheduler, adding all entries found in the platform to the labels. Collisions are resolved with comma-delimiting the values, which will act as a consistent label due to the strict ordering on properties accepted by buildbarn. Existing labels (platform, instance_name_prefix, and size_class) are prioritized, and cannot be renamed to avoid breaking existing alerts, and to avoid interfering with bb-autoscaler.

Implementation-wise, the individual metrics have been converted to templates which are instantiated whenever a new size class queue is created. The size class queue provisions and registers a registry for its metrics, which it deregisters in its `remove` method.

### DIFF
--- a/pkg/scheduler/in_memory_build_queue.go
+++ b/pkg/scheduler/in_memory_build_queue.go
@@ -46,57 +46,69 @@ import (
 var (
 	inMemoryBuildQueuePrometheusMetrics sync.Once
 
-	inMemoryBuildQueueInFlightDeduplicationsTotal = prometheus.NewCounterVec(
+	schedulerRegistry = prometheus.NewRegistry()
+
+	inMemoryBuildQueueInFlightDeduplicationsTotalTemplate = metricTemplate[prometheus.CounterOpts]{
 		prometheus.CounterOpts{
 			Namespace: "buildbarn",
 			Subsystem: "builder",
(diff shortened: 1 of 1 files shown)


pkg/scheduler/in_memory_build_queue.go
```diff
@@ -46,129 +48,192 @@ import (
...
 			Buckets:   prometheus.LinearBuckets(0, 1, 11),
 		},
-		[]string{"instance_name_prefix", "platform", "size_class"})
+		[]string{},
+	}
 )
 
+// counterTemplate is a utility struct for generating and registering prometheus.CounterVec
+// resources from a set of options, and required labels to appear in the resulting vec.
+type counterTemplate struct {
```
alexnovak, line 201: I'm wondering if there's a better place for this to live? It feels a little divorced from the rest of the logic in this file, which is already massive. Open to suggestions here.

PR #156, opened 2024-12-12
Enable directmountstrict mode for fuse

sfc-gh-guwang:
When fuse uses directmount mode, and direct mount fails, it will fall back to use 'fusermount'. BB binaries(bb-worker) doesn't have fusermount, and will always output the error message 'Fatal error: rpc error: code = Unknown desc = Failed to expose build directory mount: Failed to create FUSE server: exec: "/bin/fusermount": stat /bin/fusermount: no such file or directory'

This error message is confusing and the real error has been swallowed. 

This PR enables Buildbarn to pass along `DirectMountStrict` options that will not do fallbacks.


pkg/proto/configuration/filesystem/virtual/virtual.proto
```diff
@@ -139,6 +139,11 @@ message FUSEMountConfiguration {
   //
   // Recommended value: unset
   map<string, string> linux_backing_dev_info_tunables = 9;
+
+  // DirectMountStrict is like DirectMount but no fallback to fusermount is
+  // performed. If both DirectMount and DirectMountStrict are set,
+  // DirectMountStrict wins.
+  bool direct_mount_strict = 10;
```
EdSchouten, line 146: I'm always a big fan of making illegal states unrepresentable. There are only three behaviors:

- Always use fusermount
- Attempt to mount directly, falling back to fusermount
- Only attempt to mount directly

Yet two booleans give us four different options. Could you replace `direct_mount` and `direct_mount_strict` by an enum? Maybe something like this:

```proto
enum MountMethod {
  // Call into the fusermount utility to create the FUSE mount.
  FUSERMOUNT = 0;

  // Create the FUSE mount using the mount() system call. This can be
  // used in environments where the fusermount utility is not available,
  // such as the bb_worker container images.
  DIRECT = 1;

  // First attempt to create the FUSE mount using the mount() system
  // call. Upon failure, fall back to the fusermount utility.
  DIRECT_AND_FUSERMOUNT = 2;
}
```

sfc-gh-guwang: Updated!

pkg/filesystem/virtual/configuration/fuse_mount_enabled.go
```diff
@@ -45,6 +46,35 @@ func (m *fuseMount) Expose(terminationGroup program.Group, rootDirectory virtual
...
+
+	switch m.configuration.MountMethod {
+	case virtualpb.FUSEMountConfiguration_DIRECT:
+		mountOptions.DirectMountStrict = true
+	case virtualpb.FUSEMountConfiguration_DIRECT_AND_FUSERMOUNT:
+		mountOptions.DirectMount = true
+	case virtualpb.FUSEMountConfiguration_FUSERMOUNT:
+		// No additional options needed for FUSERMOUNT
+	default:
+		// Default to use FUSERMOUNT
```
EdSchouten, line 75: Please make the default fail:

return status.Error(codes.InvalidArgument, “invalid mount method”)

 And add an explicit case for FUSERMOUNT.

EdSchouten (commented):
This looks good. I have one final minor comment. Thanks for working on this!

EdSchouten:
Also make sure to remove the +x bit from that .pb.go file to keep CI happy.

EdSchouten:
Thanks!

PR #161, opened 2025-04-08
Update dependencies

meroton-benjamin:
Updates dependencies, in particular bb-storage, allowing use of new blob access configurations.


go.mod
```diff
@@ -28,7 +28,7 @@ require (
 	go.uber.org/mock v0.5.0
```
EdSchouten, line 28: It looks like bb-storage has not been bumped in this file. Please run a `go get -d -u ./...` or something to bump all of these.

EdSchouten (approved):

EdSchouten:
Thanks!

PR #164, opened 2025-05-01
Add support for writeable CAS backed files

meroton-benjamin:


pkg/filesystem/block_device_backed_file_pool.go
```diff
@@ -32,14 +32,35 @@ func NewBlockDeviceBackedFilePool(blockDevice blockdevice.BlockDevice, sectorAll
...
+type zeroReaderAt struct{}
+
+func (zeroReaderAt) ReadAt(p []byte, off int64) (int, error) {
+	for i := range p {
+		p[i] = 0
+	}
+	return len(p), nil
+}
+
 func (fp *blockDeviceBackedFilePool) NewFile() (filesystem.FileReadWriter, error) {
```
EdSchouten, line 44: What are your thoughts on not adding a NewFileWithReadLayer(), but simply extending the existing NewFile()?

We could provide some global instance of a zeroReaderAt that does what we want.

pkg/filesystem/read_layer.go
```diff
@@ -0,0 +1,69 @@
...
+func NewReadLayer(underlying io.ReaderAt, sizeBytes uint64) ReadLayer {
+	return &readLayer{
+		underlying: underlying,
+		sizeBytes:  sizeBytes,
+	}
+}
+
+type readLayer struct {
+	underlying io.ReaderAt
+	sizeBytes  uint64
```
EdSchouten, line 29: Probably a bit easier to use `int64` here?

pkg/filesystem/read_layer.go
```diff
@@ -0,0 +1,69 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// A ReadLayer is a read only interface for use as backing data for file
+// implementations. It behaves as a read only data source which may still be
+// shrunk (or grown) with truncate calls. Reading past the logical size returns
+// zeroes, reading at negative offsets or truncating to negative sizes is an
+// error.
+type ReadLayer interface {
```
EdSchouten, line 15: Hmmm... I'm not 100% happy with this name. I would have expected this to be named something like 'Source' or 'InitialContentsReader'. Alternatively, we could call this a 'TruncatableReaderAt'?

While there, I think that this addition of new interfaces makes FilePool a bit too big to live in pkg/filesystem. Would you be interested in moving this to a new pkg/filesystem/pool? Then it also becomes a bit more self-documenting that this interface 'belongs' to FilePool.

pkg/filesystem/read_layer.go
```diff
@@ -0,0 +1,69 @@
...
+	io.ReaderAt
+	Truncate(size int64) error
```
EdSchouten, line 17: I just realized: doesn't this need a GetNextRegionOffset() to make GetNextRegionOffset() against the pool backed file work correctly? Because whether a part of a file is sparse can now no longer be determined in isolation.

pkg/filesystem/read_layer.go
EdSchouten:
```go
type TruncatableFileReader interface {
    filesystem.FileReader
    Truncate(size int64) error
}
```

pkg/filesystem/read_layer.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func NewReadLayer(underlying io.ReaderAt, sizeBytes uint64) ReadLayer {
```
EdSchouten, line 20:
```go
func NewTruncationTrackingFileReader(filesystem.FileReader) TruncatableFileReader {
    ...
}
```

pkg/filesystem/read_layer.go
EdSchouten:
```go
type contiguousFileReader struct {
    io.ReaderAt
    length int64
}

var _ filesystem.FileReader = (*foo)(nil)

func (r * contiguousFileReader) Close() error { return nil }
func (r * contiguousFileReader) GetNextRegionOffset(...) {
        if off >= int64(len(f.data)) {
                return 0, io.EOF
        }
        switch regionType {
        case filesystem.Data:
                return off, nil
        case filesystem.Hole:
                return int64(len(f.data)), nil
        default:
                panic("Unknown region type")
        }
}
```

pkg/filesystem/block_device_backed_file_pool.go
meroton-benjamin: The latest submission modifies the interface to `NewFile(SparseReaderAt, int64)`  which accepts nil at the SparseReaderAt position.

pkg/filesystem/read_layer.go
meroton-benjamin: TruncatableSparseReaderAt uses int64

meroton-benjamin:
The latest version which properly supports GetNextRegionOffset became a bite more complicated, this is a stacked PR where the intention is for 165 and 166 to be merged ahead of this pr.

pkg/filesystem/pool/simple_sparse_reader_at.go
```diff
@@ -0,0 +1,130 @@
+package pool
```
EdSchouten, line 1: I'm not really convinced we need to ship an implementation this complex. What's the thing bb-remote-execution actually needs? Is it just an empty file? If so, why don't we offer that?

PR #165, opened 2025-05-11
Remove in_memory and directory_path file pools

meroton-benjamin:
Remove the in_memory and directory_path file pools. These filepools are better to replace with the block_device backed filepool which supports sparse files and behaves better during circumstances with scarce resources.


tomdegoede:
This PR broke any chance of Windows support where block devices are not supported. I'm experimenting with buildbarn and Bonanza and started running into this. Not sure if this was intended? @EdSchouten @meroton-benjamin

EdSchouten:
FilePool is currently only used by the virtual file system. And the virtual file system only works on Linux and macOS. So to me it's not clear how this negatively affects support for Windows. Can you please clarify?

Getting pkg/blockdevice to work on Windows would also not be very hard, especially if we only care about file backed block devices. Probably just a matter of calling [CreateFileMapping](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createfilemappinga)? I'd rather see us do that than roll back this specific change.

tomdegoede:
@EdSchouten I had a tiny POC of Buildbarn running with the native build directory and was now trying to configure Bonanza as well. I'm getting two Block device unsupported errors when starting the demo from the worker and runner. I found that the filePool settings in deployments\demo\bonanza_worker|runner.jsonnet. Therefore I was trying to change this to inMemory. I now see that filePool can be left unset in the comments, so the issue is not as big :).

PR #173, opened 2025-06-12
WIP feat(bb_scheduler): allow cancellation timouts to be configurable

tyler-french:
We deal with patterns of widespread cancellations. In these situations, it helps to be able to have the executions cancel quickly after the builds are canceled, and generally 60s is too long for this.

This PR simply adds these timeout fields to be configurable, so that when the Bazel client is killed, the execution backend is also called.


tyler-french:
Even if the RE2 was able to separate these two timeout settings, wouldn't we still want to make them configurable anyway? That way, we don't need to rebuild a container if we need to change it.

PR #178, opened 2025-07-22
Add `recursive_queued_operations_count` to BuildQueueState API

isakstenstrom:
This PR adds the field `recursive_queued_operations_count` to the `InvocationState` message in the BuildQueueState API. The field is the count of all queued operations for a invocation and all its descendants. 

With this change, it is possible to to show all queued operations for a SizeClassQueue at a time, instead of just the queued operations per invocation.

### DIFF
--- a/pkg/proto/buildqueuestate/buildqueuestate.proto
+++ b/pkg/proto/buildqueuestate/buildqueuestate.proto
@@ -235,12 +235,23 @@ message PlatformQueueState {
 }
 
 message InvocationState {
+  // TODO: This message should be used for the other counts in InvocationState.
+  message InvocationObjectCount {
+    // The number of objects that are part of this invocation.
+    uint32 direct = 1;
+
+    // The number of objects that are part of all descendants of this
+    // invocation, not including the ones that are part of this invocation.
+    uint32 indirect = 2;
+  }
+
   // Was 'id'. Moved into InvocationChildState.
   reserved 1;
 
-  // The total number of operations associated with this platform queue
-  // and invocation that are in the QUEUED execution stage.
-  uint32 queued_operations_count = 2;
+  // The number of operations associated with this platform queue
+  // and invocation that are in the QUEUED execution stage, both directly
+  // and indirectly through descendant invocations.
+  InvocationObjectCount queued_operations_count = 2;
 
   // Was 'first_queued_operation'. This information is no longer
   // provided, as the scheduler web UI no longer displays it.
--- a/pkg/scheduler/in_memory_build_queue.go
+++ b/pkg/scheduler/in_memory_build_queue.go
@@ -1850,6 +1850,15 @@ func (i *invocation) removeIfEmpty() bool {
 	return false
 }
 
+func (i *invocation) getIndirectQueuedOperationsCount() uint32 {
+	recursiveQueuedOperationsCount := uint32(0)
+	for _, iChild := range i.children {
+		recursiveQueuedOperationsCount += uint32(iChild.queuedOperations.Len())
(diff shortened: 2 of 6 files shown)


pkg/proto/buildqueuestate/buildqueuestate.proto
```diff
@@ -278,6 +278,11 @@ message InvocationState {
   // The total number of client invocations for which one or more
   // operations in the QUEUED execution stage exist.
   uint32 queued_children_count = 9;
+
+  // The total number of operations associated with this platform
+  // queue, invocation and any recursive child invocations that are in
+  // the QUEUED execution stage.
+  uint32 recursive_queued_operations_count = 10;
```
EdSchouten, line 285: Yeah, this is something that BuildQueueState is not great at. I think that for some of these counters we only report the direct value, while others are indeed recursive.

Instead of adding this additional field, what are your thoughts on having some helper type like this:

```proto
message InvocationObjectCount {
  uint32 direct = 1;
  uint32 indirect = 2;
}
```

And that we then change the type of fields like `queued_operations_count` to `InvocationObjectCount`?

pkg/scheduler/in_memory_build_queue.go
```diff
@@ -1857,14 +1859,25 @@ func (i *invocation) getInvocationState(bq *InMemoryBuildQueue) *buildqueuestate
...
 	}
+
+	recursiveQueuedOperationsCount := uint32(0)
+	children := slices.Collect(maps.Values(i.children))
+	for len(children) > 0 {
+		child := children[0]
+		children = children[1:]
+		recursiveQueuedOperationsCount += uint32(child.queuedOperations.Len())
+		children = append(children, slices.Collect(maps.Values(child.children))...)
+	}
```
EdSchouten, line 1870: Given that the tree of invocations is rather shallow, wouldn't a simple DFS implementation be better/faster/simpler?

pkg/proto/buildqueuestate/buildqueuestate.proto
```diff
@@ -235,12 +235,21 @@ message PlatformQueueState {
 }
 
 message InvocationState {
+  message InvocationObjectCount {
+    // The number of objects that are part of this invocation.
+    uint32 direct = 1;
+
+    // The number of objects that are part of this invocation and all its descendants.
```
EdSchouten, line 242: I think it would make more sense for this field to only include the count for the descendants, not the invocation itself. The advantage of doing that is that it sort of ["makes impossible states impossible"](https://kentcdodds.com/blog/make-impossible-states-impossible). Namely, if we include the value of the invocation itself, we'd also have to describe/enforce/guarantee that `direct <= indirect`.

EdSchouten (commented):
Please also make sure to also update the HTML templates under cmd/bb_scheduler to account for schema changes:
```
$ git grep -l QueuedOperationsCount cmd
cmd/bb_scheduler/templates/get_build_queue_state.html
cmd/bb_scheduler/templates/list_invocation_child_state.html
```

isakstenstrom:
> Please also make sure to also update the HTML templates under cmd/bb_scheduler to account for schema changes:
> 
> ```
> $ git grep -l QueuedOperationsCount cmd
> cmd/bb_scheduler/templates/get_build_queue_state.html
> cmd/bb_scheduler/templates/list_invocation_child_state.html
> ```

Yes! I am still working on it and will fix it later. I forgot that I was working on the same branch as the PR. I'll un-draft the PR when it is ready for another review.

isakstenstrom:
@EdSchouten Now all your feedback should be fixed.

PR #180, opened 2025-07-25
feat(scheduler): allow workers to be authorized

tomgr:
This adds authorization support to the synchronize requests, thereby allowing the scheduler to authorize workers in the usual way.

This is not a backwards-compatible change: it will require all users to specify a synchronizeAuthorizer. If you prefer I'm happy to make this backwards compatible by defaulting to an always-allow authorizer.


EdSchouten:
Thanks!

PR #184, opened 2025-08-15
Bump bb-storage

tomgr:
This also fixes a test failure on Windows.

### DIFF
--- a/cmd/bb_scheduler/BUILD.bazel
+++ b/cmd/bb_scheduler/BUILD.bazel
@@ -1,3 +1,5 @@
+load("@aspect_bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
+load("@aspect_rules_js//js:defs.bzl", "js_run_binary")
 load("@com_github_buildbarn_bb_remote_execution_npm//:purgecss/package_json.bzl", purgecss_bin = "bin")
 load("@com_github_buildbarn_bb_storage//tools:container.bzl", "container_push_official", "multiarch_go_image")
 load("@rules_go//go:def.bzl", "go_binary", "go_library")
@@ -8,8 +10,9 @@ go_library(
         "build_queue_state_service.go",
         "main.go",
     ],
+    # keep
     embedsrcs = [
-        "stylesheet.css",
+        ":stylesheet",
         "templates/error.html",
         "templates/footer.html",
         "templates/get_build_queue_state.html",
@@ -59,9 +62,11 @@ go_library(
     ],
 )
 
+templates = glob(["templates/*.html"])
+
 filegroup(
     name = "templates",
-    srcs = glob(["templates/*.html"]),
+    srcs = templates,
 )
 
 purgecss_bin.purgecss_binary(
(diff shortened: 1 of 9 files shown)


cmd/bb_scheduler/BUILD.bazel
```diff
@@ -70,16 +74,37 @@ purgecss_bin.purgecss_binary(
...
 # Create a copy of Bootstrap that only contains the style attributes
-# used by the HTML templates.
-genrule(
-    name = "stylesheet",
+# used by the HTML template.
+js_run_binary(
+    name = "stylesheet_platform",
     srcs = [
         "@com_github_twbs_bootstrap//:css/bootstrap.min.css",
-        ":templates",
```
EdSchouten, line 78: What's the reason we can't continue to use the filegroup here? Why do we need to pass in the results of `glob()` directly?

pkg/runner/local_runner_test.go
```diff
@@ -49,7 +59,7 @@ func TestLocalRunnerCheckReadiness(t *testing.T) {
 		_, err := runner.CheckReadiness(ctx, &runner_pb.CheckReadinessRequest{
 			Path: "does_not_exist",
 		})
-		testutil.RequirePrefixedStatus(t, status.Error(codes.Internal, "Failed to check existence of path \"does_not_exist\" in build directory: no such file or directory"), err)
```
EdSchouten, line 52: I'd say, let's keep it simple and change this to:
```go
testutil.RequirePrefixedStatus(t, status.Error(codes.Internal, "Failed to check existence of path \"does_not_exist\" in build directory: "), err)
```

pkg/runner/local_runner_test.go
```diff
@@ -90,7 +100,12 @@ func TestLocalRunnerRun(t *testing.T) {
 	defer buildDirectory.Close()
 
 	buildDirectoryPathBuilder, scopeWalker := path.EmptyBuilder.Join(path.VoidScopeWalker)
-	require.NoError(t, path.Resolve(path.UNIXFormat.NewParser(buildDirectoryPath), scopeWalker))
```
EdSchouten, line 93:
```go
require.NoError(t, path.Resolve(path.LocalFormat.NewParser(buildDirectoryPath), scopeWalker))
```

cmd/bb_scheduler/BUILD.bazel
tomgr: Because we can't use `locations` as we need to prefix each item in the filegroup by `../../../`. And we can only use `location` on files that are specifically mentioned in `srcs`, so even though the filegroup `templates` includes all of the items we want to call `location` on, we get an error if we do that.

EdSchouten: Gotcha! Am I correct that the `filegroup(name = "templates")` target is now unused? If so, let's remove it.

EdSchouten: Ah yes. That lint check assumes that there's always a target that has the same name as what's used in `//go:embed`. Yeah, then we should leave that.

PR #185, opened 2025-08-15
Add a winfsp-based vfs implementation

tomgr:
This change introduces a new WinFSP-based VFS implementation that goes alongside the existing FUSE and NFS VFS implementations. This enables bb_worker to succesfully execute bazel builds with virtualised file inputs with only a few drawbacks:

* Bazel really does require the Windows FS to be case insensitive, so I have had to add support for this within the VFS implementation.
* WinFSP does not support short file names and since bazel generates very long file paths, and many tools do not support long file paths (e.g. msvc), you can sometimes end up with file paths that exceed the maximum windows limit of 260. If this happens, you likely have to rename your bazel targets since WinFSP does not support short paths.

There are several forms of testing that have been done:

* There are unit tests for the WinFSP filesystem implementation that do not require WinFSP to be installed. These are quite exhaustive, with the exception of unit testing a couple of methods that are implemented by calling winfsp-provided methods.
* There are integration tests that use the winfsp-tests executable that check the filesystem behaves like a "proper windows filesystem". See https://winfsp.dev/doc/WinFsp-Tutorial/, for example. We exclude some tests from this for features we do not support, such as extended attributes.
* We have been running various versions of this in our build cluster for approximately 4 weeks with no problems.

This PR is just a draft but I'd like to get some feedback on a few specific points before I finish it, please:

1. I've vendored go-winfsp in the ffi package, as I found several issues with it and also needed to expose quite a few more functions. There are some options here:

   1. I could upstream the patches.
   2. We could use bazel to patch go-winfsp.
   3. We could stick with vendoring, in which case I will tidy up our vendored copy by, e.g., making it conform to the go style used in this repo.

   I'm happy to go with any of these.

4. I've added support for case insensitive folders for the static and in memory prepopulated directory implementations. This is a bit of an invasive change, but I cannot see another way around it.

5. I've not implemented support for removal notifications. WinFSP does have support for notifying it about file removals, but it's not clear to me that it's necessary: WinFSP itself handles removals that are initiated by it, so the only reason we would need this is if the VFS implementation (i.e. bb_worker) could delete files in parallel to an external process reading. Is that something that can happen?

6. This change introduces a new cmd, bb_virtual_dir, that just mounts a virtual file system at a particular path. This was quite helpful when developing this, particularly when debugging symlink handling. It's more of an internal testing tool though, so if you'd prefer me to remove it that's fine.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -893,6 +893,16 @@
                }
             },
             {
+               "if": "matrix.host.platform_name == 'windows_amd64'",
+               "name": "Install WinFSP",
+               "run": "choco install winfsp"
+            },
+            {
+               "if": "matrix.host.platform_name == 'windows_amd64'",
+               "name": "Execute WinFSP Integration Tests",
+               "run": "bazel test --platforms=@rules_go//go/toolchain:windows_amd64 //pkg/filesystem/virtual/winfsp:file_system_integration_test"
+            },
+            {
                "env": {
                   "GITHUB_TOKEN": "${{ secrets.GITHUB_TOKEN }}"
                },
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -107,6 +107,16 @@
                "if": "matrix.host.cross_compile || matrix.host.platform_name == 'windows_amd64'",
                "name": "windows_amd64: build${{ matrix.host.platform_name == 'windows_amd64' && ' and test' || '' }}",
                "run": "bazel ${{ matrix.host.platform_name == 'windows_amd64' && 'test --test_output=errors' || 'build' }} --platforms=@rules_go//go/toolchain:windows_amd64 //..."
+            },
+            {
+               "if": "matrix.host.platform_name == 'windows_amd64'",
+               "name": "Install WinFSP",
+               "run": "choco install winfsp"
+            },
+            {
+               "if": "matrix.host.platform_name == 'windows_amd64'",
+               "name": "Execute WinFSP Integration Tests",
+               "run": "bazel test --platforms=@rules_go//go/toolchain:windows_amd64 //pkg/filesystem/virtual/winfsp:file_system_integration_test"
             }
          ],
          "strategy": {
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -25,7 +25,7 @@ git_override(
 
 git_override(
     module_name = "com_github_buildbarn_bb_storage",
-    commit = "da347f4e972b0a2b43c1c2cc30a6835b4d941355",
+    commit = "fd11d3e1b0b435165c5e37254ac5d265e7999cb8",
     remote = "https://github.com/buildbarn/bb-storage.git",
 )
 
@@ -59,6 +59,7 @@ go_deps.from_file(go_mod = "//:go.mod")
 use_repo(
     go_deps,
     "cc_mvdan_gofumpt",
+    "com_github_aegistudio_go_winfsp",
     "com_github_bazelbuild_buildtools",
     "com_github_golang_protobuf",
     "com_github_google_uuid",
@@ -166,3 +167,16 @@ oci.pull(
     ],
 )
 use_repo(oci, "busybox", "busybox_linux_amd64", "busybox_linux_arm64_v8")
+
+http_archive(
+    name = "com_github_winfsp_winfsp_tests",
+    build_file_content = """
+filegroup(
+    name = "winfsp_tests_x64",
+    srcs = ["winfsp-tests-x64.exe"],
+    visibility = ["//visibility:public"],
+)
+""",
+    integrity = "sha256-DPxoeRcDyA+Wxynr04tjGH3lPhm6ocYDpwWdbqh8X1I=",
+    url = "https://github.com/winfsp/winfsp/releases/download/v2.1/winfsp-tests-2.1.25156.zip",
+)
--- a/cmd/bb_virtual_tmp/main.go
+++ b/cmd/bb_virtual_tmp/main.go
@@ -59,14 +59,15 @@ func main() {
 			"bb_virtual_tmp",
 			/* rootDirectory = */ virtual_configuration.LongAttributeCaching,
 			/* childDirectories = */ virtual_configuration.LongAttributeCaching,
-			/* leaves = */ virtual_configuration.NoAttributeCaching)
+			/* leaves = */ virtual_configuration.NoAttributeCaching,
+			/* caseSensitive = */ true)
 		if err != nil {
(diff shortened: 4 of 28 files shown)


EdSchouten:
Hey Thomas,

Exciting stuff!

> Bazel really does require the Windows FS to be case insensitive, so I have had to add support for this within the VFS implementation.

What are your thoughts on adding an interface like this?

```go
type NormalizedComponent string

type ComponentNormalizer interface {
    Normalize(path.Component) string
}
```

And that we then add implementations like these:

```go
type caseSensitiveComponentNormalizer struct {}

func (caseSensitiveComponentNormalizer) Normalize(c path.Component) string {
    return NormalizedComponent(c.String())
}

var CaseSensitiveComponentNormalizer ComponentNormalizer = caseSensitiveComponentNormalizer{}

// And another one that calls strings.ToLower()
```

macOS is case insensitive by default, but we currently let workers be case sensitive. I think this is actually desirable for most users, but I can imagine that some of them may need case insensitivity as well. Then there's also the question about what to do with Unicode normalization (NFC vs NFD), so letting the algorithm be pluggable may be of use.

Also be sure to make the configuration option for controlling case sensitive be independent of WinFSP. Just put it next to `shuffle_directory_listings` in `VirtualBuildDirectoryConfiguration`.

> WinFSP does not support short file names and since bazel generates very long file paths, and many tools do not support long file paths (e.g. msvc), you can sometimes end up with file paths that exceed the maximum windows limit of 260. If this happens, you likely have to rename your bazel targets since WinFSP does not support short paths.

Just to check: WinFSP **only supports** short file names, right. That's good to know. I suppose that it won't be a big problem in practice, because users can give the build directory its own drive letter. That should cut down the path length a lot.

> I've vendored go-winfsp in the ffi package, as I found several issues with it and also needed to expose quite a few more functions. There are some options here:

My recommendation would be to go for (1). And while we're waiting for those changes to be merged, do (2). I'm not a big fan of vendoring this library into our tree, also because its license is different from that of Buildbarn.

> I've not implemented support for removal notifications. WinFSP does have support for notifying it about file removals, but it's not clear to me that it's necessary: WinFSP itself handles removals that are initiated by it, so the only reason we would need this is if the VFS implementation (i.e. bb_worker) could delete files in parallel to an external process reading. Is that something that can happen?

Don't worry about that. Notifications are mostly just needed by bb_clientd's in combination with Bazel's `--remote_output_service`. There we have bb_clientd and Bazel mutating the same file system in parallel, which requires these notifications. Workers don't mutate the build directory while an action is running. We also don't implement it for NFSv4 and there things work fine.

> This change introduces a new cmd, bb_virtual_dir, that just mounts a virtual file system at a particular path. This was quite helpful when developing this, particularly when debugging symlink handling. It's more of an internal testing tool though, so if you'd prefer me to remove it that's fine.

Considering that bb_clientd also offers a `/scratch`, I'd argue that we don't really need it. Even if you want to hack on the VFS code and test stuff, you can always build a version of bb_clientd with local VFS changes by using `--override_module=...`.

tools/constraints/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
```
EdSchouten, line 1: Instead of adding this build setting, what are your thoughts on simply marking the test as `tags = ["manual"]`? Then we can add a separate CI step that only runs on Windows calling `bazel test //the/integration/test`. I think this is acceptable, given that we already have a CI step that installs WinFSP.

EdSchouten: Sure!

pkg/filesystem/virtual/winfsp/winfsp_windows.go
```diff
@@ -0,0 +1,1424 @@
...
+
+func NewWinFSPFileSystem(rootDirectory virtual.Directory, caseSensitive bool) *WinfspFileSystem {
+	return &WinfspFileSystem{
+		caseSensitive: caseSensitive,
+		rootDirectory: rootDirectory,
+		openFiles:     newOpenedNodesPool(),
+	}
+}
+
+func (fs *WinfspFileSystem) Mount(mountpoint string) (*ffi.FileSystem, error) {
```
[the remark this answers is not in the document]
tomgr: I think I have to make WinfspFileSystem a public type in order to for the unit testing in winfsp_windows_test.go to work as those directly invoke the WinfspFileSystem methods directly.

Originally this method didn't exist and it was part of NewWinFSPFileSystem, but I split them when adding the unit tests as those need to create the struct, but can't mount it as WinFSP might not be installed. I'm happy to move it if you prefer; the reason I put it here is because it encapsulates all of the interaction with go-winfsp into this file, and also keeps all of the filesystem settings close to the implementations.

pkg/filesystem/virtual/winfsp/opened_files_pool_windows.go
```diff
@@ -0,0 +1,103 @@
...
+	"sync"
+
+	"github.com/buildbarn/bb-remote-execution/pkg/filesystem/virtual"
+	path "github.com/buildbarn/bb-storage/pkg/filesystem/path"
+
+	"golang.org/x/sys/windows"
+)
+
+// openedNodesPool holds the state of nodes that are opened by WinFSP.
+type openedNodesPool struct {
```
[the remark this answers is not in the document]
tomgr: Nice to know - thanks! I think I'll leave it as-is as I quite like at least some separation of some bookkeeping code.

tools/constraints/BUILD.bazel
EdSchouten: That's slick!

tomgr:
Thank for the comments Ed. Think I've addressed them all so it should be ready for another review now.

moroten:
Impressive work! When all this has propagated through to `bb-deployments`, would it be possible to add integration tests there for a pure Windows deployment? Do you see any blockers for that? I'm thinking about a "bare" deployment test.

tomgr:
Thanks! No I don't see any blockers as github actions supports easy installation of WinFSP (`choco install winfsp`) and WinFSP does work fine on gha as there is an integration test here that does exactly that. I guess the only thing that will require some care is making the configs compatible with Windows which is mostly just a case of making sure the configs use Windows-style paths in various places.

I'd be happy to have a look at that if you like. Were you thinking of a Windows version of `test-deployment-bare.sh`?

moroten:
> I'd be happy to have a look at that if you like. Were you thinking of a Windows version of `test-deployment-bare.sh`?

Yes, exactly. Having at least one integration test and example would be great! (I have no experience with Docker on Windows, but all contributions are welcome.)

moroten:
@tomgr bb-deployments has now been updated to include this commit.

PR #186, opened 2025-08-19
Update bb-storage and implement NewTicker

tomgr:
This provides a simple implementation of NewTicker in terms of NewTimer. Testing this is a bit of a pain but these tests appear to pass robustly on my machine.


tomgr:
Ahh I see you've worked around this - up to you if you want this implementation or not. It's not that sophisticated :)

EdSchouten:
Yeah, I only added a stub to keep the build working. Let's use your implementation. LGTM! Will merge it as soon as you address the CI failure.

PR #187, opened 2025-08-21
Workaround MSVC header inclusion issues

tomgr:
This works-around this bug in bazel: https://github.com/bazelbuild/bazel/issues/19733. In short, Bazel's C++ header inclusion checking is broken when executing MSVC under remote execution. This commit works-around this issue for bzlmod-based builds following a suggestion in that thread by making all builds run inside a specific named subfolder. We thus change the input root to `execroot/_main` on Windows, instead of the previous default of `root`.


moroten:
The header inclusion check is only interesting if you want to use [unused_input_list](https://bazel.build/rules/lib/builtins/actions#run) for reducing rebuilds when "unused" input files have changed. I think that feature feels scary and would rather disable it unless you see production loss at your company.

PR #188, opened 2025-09-03
Allow configuration of build directory owner

mortenmj:
This allows us to set the owner UID/GID for the build directory, when using a virtual file system

### DIFF
--- a/cmd/bb_worker/main.go
+++ b/cmd/bb_worker/main.go
@@ -372,7 +372,9 @@ func main() {
 								suspendableClock),
 							symlinkFactory,
 							characterDeviceFactory,
-							handleAllocator)
+							handleAllocator,
+							runnerConfiguration.BuildDirectoryOwnerUserId,
+							runnerConfiguration.BuildDirectoryOwnerGroupId)
 					} else {
 						executionTimeoutClock = clock.SystemClock
 						buildDirectory = builder.NewNaiveBuildDirectory(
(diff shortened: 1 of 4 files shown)


cmd/bb_worker/main.go
```diff
@@ -236,9 +236,8 @@ func main() {
 					clock.SystemClock,
 					normalizer,
 					/* defaultAttributesSetter = */ func(requested virtual.AttributesMask, attributes *virtual.Attributes) {
-						// No need to set ownership
```
EdSchouten, line 239: Wait. This affects the ownership on the top-level build directory. Not the per-action directory. This should be left as is. We should change pkg/builder/virtual_build_directory.go instead.

EdSchouten (changes requested):

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -299,6 +299,16 @@ message VirtualBuildDirectoryConfiguration {
   // bazel on Windows effectively assumes that the file system is case
   // insensitive, so make sure this is set to true when using the WinFSP.
   bool case_insensitive = 6;
+
+  // The ID of the user that owns directories contained in the build
+  // directory. This should typically be set to the user that is used by
+  // bb_runner to run actions.
+  uint32 build_directory_owner_user_id = 7;
```
EdSchouten, line 306: Ideally these options live inside RunnerConfiguration. The reason being that it's not uncommon to run a single worker and multiple runners. The latter having a distinct user ID each.

PR #190, opened 2025-09-21
Let filepool files be owned by user

meroton-benjamin:
```
This commit extends the `buildDirectoryOwnerGroupId` and
`buildDirectoryOwnerUserId` fields to also apply to filepool backed
files. CAS-backed files remains owned by the root user.
```

### DIFF
--- a/pkg/filesystem/virtual/pool_backed_file_allocator.go
+++ b/pkg/filesystem/virtual/pool_backed_file_allocator.go
@@ -45,8 +45,10 @@ var (
 )
 
 type poolBackedFileAllocator struct {
-	pool        pool.FilePool
-	errorLogger util.ErrorLogger
+	pool         pool.FilePool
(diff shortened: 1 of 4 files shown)


pkg/filesystem/virtual/pool_backed_file_allocator.go
```diff
@@ -332,6 +342,8 @@ func (f *fileBackedFile) virtualGetAttributesLocked(attributes *Attributes) {
 	}
 	attributes.SetPermissions(permissions)
 	attributes.SetSizeBytes(f.size)
+	attributes.SetOwnerUserID(f.ownerUserID)
+	attributes.SetOwnerGroupID(f.ownerGroupID)
```
EdSchouten, line 346: You can omit this. All calls to `virtualGetAttributesLocked()` are preceded by `virtualGetAttributesUnlocked()`.

PR #200, opened 2025-10-24
pkg/filesystem: Make the zeroes underlying a pool file changeable

Jille:
This allows bonanza to have (mutable) files backed by others. Bonanza
can use another FileReader to supply the data for all the "holes" in a
pool file.

Note that this requires some care to work correctly: If a file is
truncated to shrink and later extended again, the HoleSource must track
that and emit zeroes from offset of the smallest truncation.

### DIFF
--- a/internal/mock/BUILD.bazel
+++ b/internal/mock/BUILD.bazel
@@ -183,6 +183,7 @@ gomock(
     out = "filesystem_filepool.go",
     interfaces = [
         "FilePool",
+        "HoleSource",
         "SectorAllocator",
     ],
     library = "//pkg/filesystem/pool",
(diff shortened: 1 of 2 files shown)


pkg/filesystem/pool/block_device_backed_file_pool.go
```diff
@@ -129,36 +134,43 @@ func (f *blockDeviceBackedFile) GetNextRegionOffset(off int64, regionType filesy
...
 			sectorIndex++
-			if f.sectors[sectorIndex] != 0 {
-				return int64(sectorIndex) * sectorSizeBytes, nil
-			}
 		}
-	case filesystem.Hole:
-		if sectorIndex >= len(f.sectors) || f.sectors[sectorIndex] == 0 {
-			// Already inside a hole.
-			return off, nil
+		o := int64(sectorIndex) * sectorSizeBytes
```
EdSchouten: Maybe rename `o` to `fileNextRegionOffset and `u` to `holeSourceNextRegionOffset`?

PR #210, opened 2026-02-02
Allow Darwin NFS mounts to be exposed on TCP ports

tomgr:
The main reason for wanting this was to allow running bb_runner inside VMs whilst bb_worker is on the main host.


mikaoura:
I'm running into some trouble trying to get the unix socket -based nfs working on macOS Tahoe. The mount system call from bb_clientd is failing. I'll dig deeper, but it might also be a motivation to supporting TCP. The macOS documentation only talks about TCP support, no mention of unix sockets.

PR #222, opened 2026-05-03
fuse: configurable MaxBackground from go-fuse, minor cleanups

aron-muon:
- expose maximum_background in FUSEMountConfiguration

  - Adds a `maximum_background` field to `FUSEMountConfiguration`. When set, it caps the FUSE asynchronous request queue (FUSE_INIT `max_background`); when zero, go-fuse's default of 12 is used.

  - The default of 12 suits `bb_clientd`, but is too low for `bb_worker`, where a single mount serves many concurrent actions. The kernel `CongestionThreshold` is `max_background * 3/4`, and once it's reached, reads beyond it queue in D-state, visible as the worker wedging while continuing to report actions as executing.

  - Our recommended value for `bb_worker` is `1024`. Validated in production at Muon Space against full `//...` Rust builds.

- bump bb-storage to upstream/main HEAD

  - Picks up [buildbarn/bb-storage#332](https://github.com/buildbarn/bb-storage/pull/332) ("Forward termination signals to child process"). Without it, processes running as PID 1 in containers ignore SIGTERM and the kubelet has to fall back to SIGKILL once the grace period expires. For `bb_worker` that means in-flight actions get killed and have to be re-executed elsewhere. Bumps both `go.mod` (gazelle's `go_deps` source) and the `git_override` in `MODULE.bazel` that controls bzlmod resolution.

- load proto_library from @protobuf//bazel

  - Routine cleanup: `@rules_proto//proto:defs.bzl` is deprecated in favour of `@protobuf//bazel:proto_library.bzl` in newer `protobuf` module versions. Same change across all 14 `pkg/proto/**/BUILD.bazel` files.


pkg/filesystem/virtual/configuration/fuse_mount_enabled.go
```diff
@@ -65,6 +65,16 @@ func (m *fuseMount) Expose(terminationGroup program.Group, rootDirectory virtual
...
 		EnableWritebackCache: true,
+		// Override go-fuse's default of 12 — that limit caps the kernel-side
+		// FUSE async-request queue at 12 in-flight operations per mount, with
+		// CongestionThreshold = MaxBackground * 3/4 = 9. With ≥8 concurrent
+		// actions per worker each doing readahead, the queue saturates almost
+		// instantly and subsequent reads block in kernel D-state, leading to
+		// the well-known wedge under high concurrency. 1024 is generous
+		// headroom for typical worker concurrency × per-action open files.
+		// TODO: expose this via FUSEMountConfiguration proto in a follow-up
+		// PR upstream; hardcoded for the staging RE rollout in May 2026.
```
EdSchouten: SGTM w.r.t. this approach.

aron-muon: Ready for review, this approach is implemented (along with bumping dep of bb storage to pick up fixes)

pkg/proto/configuration/filesystem/virtual/virtual.proto
```diff
@@ -171,6 +171,20 @@ message FUSEMountConfiguration {
...
+  // the max_background field in the FUSE_INIT reply. The kernel
+  // additionally treats max_background * 3/4 as a congestion threshold;
+  // requests beyond it block in the kernel until the queue drains.
+  //
+  // The go-fuse default of 12 is suitable for bb_clientd and similar
+  // low-concurrency consumers. For bb_worker, where a single mount
+  // serves many concurrent actions and each may issue readahead, the
+  // queue saturates quickly and reads stall in the kernel.
+  //
+  // Recommended value: 1024 for bb_worker, unset for other consumers.
```
EdSchouten: My gut feeling tells me that the recommended value should be proportional to the number of worker threads, instead of a constant value. Probably good to document it as such?

pkg/proto/configuration/filesystem/virtual/virtual.proto
```diff
@@ -171,6 +171,20 @@ message FUSEMountConfiguration {
...
+  uint32 maximum_background = 11;
```
EdSchouten: I'm all for keeping the naming of these fields consistent with go-fuse as much as possible. However, I don't think this name is very descriptive. Given that the default value of this option has a more sensible name (`_DEFAULT_BACKGROUND_TASKS`), what are your thoughts on calling this `maximum_background_tasks`?

EdSchouten (commented):
Hi Aron,
LGTM, though I have a couple of tiny remarks. Let me know what you think!

EdSchouten (approved):

pkg/proto/configuration/filesystem/virtual/virtual.proto
aron-muon: Renamed to `maximum_background_tasks`!

pkg/proto/configuration/filesystem/virtual/virtual.proto
aron-muon: Updated doc with that context - will be valuable for future readers

PR #225, opened 2026-05-19
feature: new image router

finn-ball:


EdSchouten:
We intentionally added support for the remote action router, so that features like these do not need to be made part of the scheduler.

EdSchouten:
Even though I am convinced all of this works fine and achieves all of the stated goals, my concern is that this code as a result of being AI generated does not resemble the style and patterns that you see in the rest of the code base. I am therefore pretty hesitant towards merging it into this repository.

That said, just like with bb-remote-asset, bb-portal, etc. I am fine with giving you a separate repository in this organization that you can use to host this code. Would that work for you?

finn-ball:
> Even though I am convinced all of this works fine and achieves all of the stated goals, my concern is that this code as a result of being AI generated does not resemble the style and patterns that you see in the rest of the code base. I am therefore pretty hesitant towards merging it into this repository.
> 
> That said, just like with bb-remote-asset, bb-portal, etc. I am fine with giving you a separate repository in this organization that you can use to host this code. Would that work for you?

I would rather have this merged into the main codebase so it can enjoy proper maintenance as I think this is a useful service to add. It would tie in well with mounting special filesystems into the `chroot` worker too, which I am very keen to look at after this work has been complete.

I'll rewrite it myself and try to stick more to your coding guidelines but it may take me a bit longer to do so. Let me know if that doesn't work for you and you still want this to be an entirely separate repository of work :)

finn-ball:
https://github.com/buildbarn/bb-action-router

Closing in favour of this

PR #226, opened 2026-05-22
Enhancements made for nfs serving

visualphoenix:
Hi @EdSchouten ❤️ 

Not sure what you'll think about these changes, but fwiw I successfully used these patches to pass 100% of pjdfstest on a pure go nfs v4/v4.1 server I've been hacking on.

I understand if they aren't what you want upstream, but I figured I'd put them up for your consideration.

Thank you again for maintaining this wonderful code base.

### DIFF
--- a/pkg/filesystem/virtual/attributes.go
+++ b/pkg/filesystem/virtual/attributes.go
@@ -37,9 +37,15 @@ const (
 	// This attribute only needs to be provided by regular files and
 	// directories, as NFSv4 uses distinct file types for those.
 	AttributesMaskIsInNamedAttributeDirectory
+	// AttributesMaskLastAccessTime requests the last access time
+	// (st_atim).
+	AttributesMaskLastAccessTime
 	// AttributesMaskLastDataModificationTime requests the last data
 	// modification time (st_mtim).
 	AttributesMaskLastDataModificationTime
+	// AttributesMaskLastStatusChangeTime requests the last status
+	// change time (st_ctim).
+	AttributesMaskLastStatusChangeTime
 	// AttributesMaskLinkCount requests the link count (st_nlink).
 	AttributesMaskLinkCount
 	// AttributesMaskOwnerGroupID requests the ID of the group that
@@ -72,7 +78,9 @@ type Attributes struct {
 	hasNamedAttributes              bool
 	inodeNumber                     uint64
 	isInsideNamedAttributeDirectory bool
+	lastAccessTime                  time.Time
 	lastDataModificationTime        time.Time
+	lastStatusChangeTime            time.Time
 	linkCount                       uint32
 	ownerGroupID                    uint32
 	ownerUserID                     uint32
@@ -195,6 +203,18 @@ func (a *Attributes) SetIsInNamedAttributeDirectory(isInsideNamedAttributeDirect
 	return a
 }
 
+// GetLastAccessTime returns the last access time (st_atim).
+func (a *Attributes) GetLastAccessTime() (time.Time, bool) {
+	return a.lastAccessTime, a.fieldsPresent&AttributesMaskLastAccessTime != 0
+}
+
+// SetLastAccessTime sets the last access time (st_atim).
+func (a *Attributes) SetLastAccessTime(lastAccessTime time.Time) *Attributes {
+	a.lastAccessTime = lastAccessTime
+	a.fieldsPresent |= AttributesMaskLastAccessTime
+	return a
+}
+
 // GetLastDataModificationTime returns the last data modification time
 // (st_mtim).
 func (a *Attributes) GetLastDataModificationTime() (time.Time, bool) {
@@ -206,6 +226,18 @@ func (a *Attributes) GetLastDataModificationTime() (time.Time, bool) {
 func (a *Attributes) SetLastDataModificationTime(lastDataModificationTime time.Time) *Attributes {
 	a.lastDataModificationTime = lastDataModificationTime
 	a.fieldsPresent |= AttributesMaskLastDataModificationTime
+	return a
+}
+
+// GetLastStatusChangeTime returns the last status change time (st_ctim).
+func (a *Attributes) GetLastStatusChangeTime() (time.Time, bool) {
+	return a.lastStatusChangeTime, a.fieldsPresent&AttributesMaskLastStatusChangeTime != 0
+}
+
+// SetLastStatusChangeTime sets the last status change time (st_ctim).
+func (a *Attributes) SetLastStatusChangeTime(lastStatusChangeTime time.Time) *Attributes {
+	a.lastStatusChangeTime = lastStatusChangeTime
+	a.fieldsPresent |= AttributesMaskLastStatusChangeTime
 	return a
 }
 
--- a/pkg/filesystem/virtual/fuse/simple_raw_file_system.go
+++ b/pkg/filesystem/virtual/fuse/simple_raw_file_system.go
@@ -54,6 +54,8 @@ func toFUSEStatus(s virtual.Status) fuse.Status {
 		return fuse.EIO
 	case virtual.StatusErrIsDir:
 		return fuse.EISDIR
+	case virtual.StatusErrNameTooLong:
+		return fuse.Status(syscall.ENAMETOOLONG)
 	case virtual.StatusErrNoEnt:
 		return fuse.ENOENT
 	case virtual.StatusErrNotDir:
--- a/pkg/filesystem/virtual/fuse/simple_raw_file_system.go
+++ b/pkg/filesystem/virtual/fuse/simple_raw_file_system.go
@@ -433,6 +433,7 @@ func (rfs *simpleRawFileSystem) Mknod(cancel <-chan struct{}, input *fuse.MknodI
 
 	var createAttributes virtual.Attributes
 	createAttributes.SetFileType(fileType)
+	createAttributes.SetPermissions(virtual.NewPermissionsFromMode(input.Mode & 0o7777))
 	var attributes virtual.Attributes
 	child, _, vs := i.VirtualMknod(ctx, path.MustNewComponent(name), &createAttributes, AttributesMaskForFUSEAttr, &attributes)
 	if vs != virtual.StatusOK {
@@ -451,8 +452,10 @@ func (rfs *simpleRawFileSystem) Mkdir(cancel <-chan struct{}, input *fuse.MkdirI
 	i := rfs.getDirectoryLocked(input.NodeId)
 	rfs.nodeLock.RUnlock()
 
+	var createAttributes virtual.Attributes
+	createAttributes.SetPermissions(virtual.NewPermissionsFromMode(input.Mode & 0o7777))
 	var attributes virtual.Attributes
-	child, _, vs := i.VirtualMkdir(ctx, path.MustNewComponent(name), AttributesMaskForFUSEAttr, &attributes)
+	child, _, vs := i.VirtualMkdir(ctx, path.MustNewComponent(name), &createAttributes, AttributesMaskForFUSEAttr, &attributes)
 	if vs != virtual.StatusOK {
 		return toFUSEStatus(vs)
 	}
--- a/pkg/filesystem/virtual/fuse/simple_raw_file_system.go
+++ b/pkg/filesystem/virtual/fuse/simple_raw_file_system.go
@@ -438,45 +438,61 @@ func (rfs *simpleRawFileSystem) Mknod(cancel <-chan struct{}, input *fuse.MknodI
 }
 
 func (rfs *simpleRawFileSystem) Mkdir(cancel <-chan struct{}, input *fuse.MkdirIn, name string, out *fuse.EntryOut) fuse.Status {
+	ctx, s := rfs.createContext(cancel, &input.Caller)
+	if s != fuse.OK {
+		return s
+	}
 	rfs.nodeLock.RLock()
 	i := rfs.getDirectoryLocked(input.NodeId)
 	rfs.nodeLock.RUnlock()
 
 	var attributes virtual.Attributes
-	child, _, s := i.VirtualMkdir(path.MustNewComponent(name), AttributesMaskForFUSEAttr, &attributes)
-	if s != virtual.StatusOK {
-		return toFUSEStatus(s)
+	child, _, vs := i.VirtualMkdir(ctx, path.MustNewComponent(name), AttributesMaskForFUSEAttr, &attributes)
+	if vs != virtual.StatusOK {
+		return toFUSEStatus(vs)
 	}
 	rfs.addDirectory(child, &attributes, out)
 	return fuse.OK
 }
 
 func (rfs *simpleRawFileSystem) Unlink(cancel <-chan struct{}, header *fuse.InHeader, name string) fuse.Status {
+	ctx, s := rfs.createContext(cancel, &header.Caller)
+	if s != fuse.OK {
+		return s
+	}
 	rfs.nodeLock.RLock()
 	i := rfs.getDirectoryLocked(header.NodeId)
 	rfs.nodeLock.RUnlock()
 
-	_, s := i.VirtualRemove(path.MustNewComponent(name), false, true)
-	return toFUSEStatus(s)
+	_, vs := i.VirtualRemove(ctx, path.MustNewComponent(name), false, true)
+	return toFUSEStatus(vs)
 }
 
 func (rfs *simpleRawFileSystem) Rmdir(cancel <-chan struct{}, header *fuse.InHeader, name string) fuse.Status {
+	ctx, s := rfs.createContext(cancel, &header.Caller)
+	if s != fuse.OK {
+		return s
+	}
 	rfs.nodeLock.RLock()
 	i := rfs.getDirectoryLocked(header.NodeId)
 	rfs.nodeLock.RUnlock()
 
-	_, s := i.VirtualRemove(path.MustNewComponent(name), true, false)
-	return toFUSEStatus(s)
+	_, vs := i.VirtualRemove(ctx, path.MustNewComponent(name), true, false)
+	return toFUSEStatus(vs)
 }
 
 func (rfs *simpleRawFileSystem) Rename(cancel <-chan struct{}, input *fuse.RenameIn, oldName, newName string) fuse.Status {
+	ctx, s := rfs.createContext(cancel, &input.Caller)
+	if s != fuse.OK {
+		return s
+	}
 	rfs.nodeLock.RLock()
 	iOld := rfs.getDirectoryLocked(input.NodeId)
 	iNew := rfs.getDirectoryLocked(input.Newdir)
 	rfs.nodeLock.RUnlock()
 
-	_, _, s := iOld.VirtualRename(path.MustNewComponent(oldName), iNew, path.MustNewComponent(newName))
-	return toFUSEStatus(s)
+	_, _, vs := iOld.VirtualRename(ctx, path.MustNewComponent(oldName), iNew, path.MustNewComponent(newName))
+	return toFUSEStatus(vs)
 }
 
 func (rfs *simpleRawFileSystem) Link(cancel <-chan struct{}, input *fuse.LinkIn, filename string, out *fuse.EntryOut) fuse.Status {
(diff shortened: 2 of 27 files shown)


pkg/filesystem/virtual/nfsv4/nfs40_program.go
```diff
@@ -2473,6 +2473,8 @@ func toNFSv4Status(s virtual.Status) nfsv4.Nfsstat4 {
 		return nfsv4.NFS4ERR_WRONG_TYPE
 	case virtual.StatusErrXDev:
 		return nfsv4.NFS4ERR_XDEV
+	case virtual.StatusErrNameTooLong:
+		return nfsv4.NFS4ERR_NAMETOOLONG
```
EdSchouten, line 2507: Could you also patch up the FUSE and WinFSP backends to return the right status codes? So extend:

- pkg/filesystem/virtual/fuse/simple_raw_file_system.go toFUSEStatus()
  - You probably need to use `fuse.Status(syscall.ENAMETOOLONG)` here, as go-fuse doesn't have a constant for this.
- pkg/filesystem/virtual/winfsp/file_system.go toNTStatus()
  - Probably `STATUS_NAME_TOO_LONG`? Not sure.

pkg/filesystem/virtual/fuse/simple_raw_file_system.go
```diff
@@ -441,7 +441,7 @@ func (rfs *simpleRawFileSystem) Mkdir(cancel <-chan struct{}, input *fuse.MkdirI
 	rfs.nodeLock.RUnlock()
 
 	var attributes virtual.Attributes
-	child, _, s := i.VirtualMkdir(path.MustNewComponent(name), AttributesMaskForFUSEAttr, &attributes)
+	child, _, s := i.VirtualMkdir(context.Background(), path.MustNewComponent(name), AttributesMaskForFUSEAttr, &attributes)
```
EdSchouten: Instead of passing in `context.Background()` here, you should be able to do this to get a context:
```go
	ctx, s := rfs.createContext(cancel, &input.Caller)
	if s != fuse.OK {
		return s
	}
```

pkg/filesystem/virtual/winfsp/file_system.go
```diff
@@ -536,7 +536,7 @@ func (FileSystem) openOrCreateDir(ctx context.Context, parent virtual.Directory,
 	}
 
 	// Create new directory.
-	childDir, _, status := parent.VirtualMkdir(leafName, AttributesMaskForWinFSPAttr, attributes)
+	childDir, _, status := parent.VirtualMkdir(context.Background(), leafName, AttributesMaskForWinFSPAttr, attributes)
```
EdSchouten: It looks like these methods already have a `ctx`. You can probably just use that one.

pkg/filesystem/virtual/leaf.go
EdSchouten: I guess that `VirtualAllocate()`, `VirtualSeek()`, and `VirtualRead()` are now the only methods that still need `ctx`. Feel free to also patch those up and remove this `TODO`. Or leave the code as it is now. Both are fine by me! :-)

pkg/filesystem/virtual/nfsv4/nfs41_program.go
```diff
@@ -2669,13 +2672,11 @@ func (s *sequenceState) opSecInfoNoName(args *nfsv4.SecinfoNoName4args) nfsv4.Se
 	}
 
 	s.currentFileHandle = nfs41FileHandle{}
-	return &nfsv4.Secinfo4res_NFS4_OK{
-		Resok4: []nfsv4.Secinfo4{
-			&nfsv4.Secinfo4_default{
-				Flavor: rpcv2.AUTH_NONE,
-			},
-		},
+	flavors := make([]nfsv4.Secinfo4, 0, len(s.program.securityFlavors))
```
EdSchouten: Instead of constructing this slice, what are your thoughts on letting `NewNFS4[01]Program()` take a `[]nfsv4.Secinfo4`?

pkg/filesystem/virtual/nfsv4/nfs41_program.go
```diff
@@ -74,7 +75,13 @@ func NewNFS41Program(
 	clock clock.Clock,
 	enforcedLeaseTime, announcedLeaseTime time.Duration,
 	pathFormat path.Format,
+	securityFlavors []rpcv2.AuthFlavor,
 ) nfsv4.Nfs4Program {
+	if len(securityFlavors) == 0 {
+		// Backwards compatibility: bb's original loopback consumer
+		// expected SECINFO/SECINFO_NO_NAME to advertise AUTH_NONE only.
+		securityFlavors = []rpcv2.AuthFlavor{rpcv2.AUTH_NONE}
+	}
```
EdSchouten: I'd say, let's not add this specific piece of code. It should be up to the caller of `NewNFS4[01]Program()` to pass in a proper list.

pkg/filesystem/virtual/attributes.go
```diff
@@ -64,6 +64,17 @@ const (
...
+	// bits of st_mode), including the setuid, setgid, and sticky
+	// bits in addition to the rwxrwxrwx access bits.
+	//
+	// AttributesMaskPermissions captures only the union of the
+	// owner/group/other access bits and is sufficient for protocol
+	// access checks; AttributesMaskMode preserves the full 12-bit
+	// mode so chmod-style operations round-trip faithfully through
+	// callers that care (e.g., a host-FS-backed NFS server that
+	// must report setuid/setgid/sticky to clients).
+	AttributesMaskMode
```
EdSchouten: Instead of making this specific change right now, can you please remove it from this PR and file an issue? I think there is some larger work we need to do across all of Buildbarn to tidy up/unify the way we select file permissions. @meroton-benjamin @moroten  and I also had discussions in the past about how the permissions of files can be improved. Thanks!

pkg/filesystem/virtual/nfsv4/nfs40_program.go
```diff
@@ -3322,8 +3322,35 @@ func fattr4ToAttributes(in *nfsv4.Fattr4, out *virtual.Attributes) nfsv4.Nfsstat
...
+			// See FATTR4_OWNER above for the numeric-only rationale.
+			group, _, err := nfsv4.ReadUtf8strMixed(r)
+			if err != nil {
+				return nfsv4.NFS4ERR_BADXDR
+			}
+			gid, err := strconv.ParseUint(string(group), 10, 32)
+			if err != nil {
+				return nfsv4.NFS4ERR_BADOWNER
+			}
+			out.SetOwnerGroupID(uint32(gid))
```
EdSchouten, line 3381: If we're going to make this change, then we should make sure that all existing implementations of `VirtualSetAttributes()` return `StatusErrPerm` when either of those options is set.

Also, pkg/filesystem/virtual/fuse/simple_raw_file_system.go's `SetAttr()` could use similar treatment.

pkg/filesystem/virtual/attributes.go
```diff
@@ -90,7 +125,9 @@ type Attributes struct {
 	inodeNumber                     uint64
 	isInsideNamedAttributeDirectory bool
 	lastAccessTime                  time.Time
+	lastAccessTimeForSet            SettableTime
 	lastDataModificationTime        time.Time
+	lastDataModificationTimeForSet  SettableTime
```
EdSchouten: I'm just thinking what we can do to support this feature in a way that's as light-weight/streamlined as possible.

What if, instead of having a separate field and a `SettableTime` we reuse the existing AttributeMask values and Attributes fields, and do either one of the following two things:

- Document that if `Attributes.lastAccessTime.IsZero()`, the value should be interpreted as the current time of day,
- Let NFSv4[01]Program call `clock.Now()` to whenever the client requests that the time on the server is used? Probably not 100% accurate, but maybe good enough for what you're trying to support?

Would that be sufficient for your use case?

pkg/filesystem/virtual/nfsv4/nfs40_program.go
```diff
@@ -1056,16 +1074,23 @@ func (s *compoundState) opCreate(ctx context.Context, args *nfsv4.Create4args) n
...
 	var vs virtual.Status
 	switch objectType := args.Objtype.(type) {
-	case *nfsv4.Createtype4_NF4BLK, *nfsv4.Createtype4_NF4CHR:
-		// Character and block devices can only be provided as
-		// part of input roots, if workers are set up to provide
-		// them. They can't be created through the virtual file
-		// system.
-		return &nfsv4.Create4res_default{Status: nfsv4.NFS4ERR_PERM}
+	case *nfsv4.Createtype4_NF4BLK:
+		attributes.SetDeviceNumber(filesystem.NewDeviceNumberFromMajorMinor(objectType.Devdata.Specdata1, objectType.Devdata.Specdata2))
```
EdSchouten: I don't think that this is correct/in the spirit of how `VirtualMknod()` was designed. Right now it takes a FileType as an input, and a virtual.Attributes as its output. So we should always be passing a zeroed instance of `virtual.Attributes` to `VirtualMknod()`.

If we wanted to support creating block/character devices, my recommendation would be to rework `VirtualMknod()` to something like this:

```
VirtualMknod(ctx context.Context, name path.Component, createAttributes *virtual.Attributes, requested AttributesMask, createdFileAttributes *Attributes) (Leaf, ChangeInfo, Status)
```

So take two disjoint instances of `virtual.Attributes`. This is similar to what we do with `VirtualOpenChild()`.

I'd also like to see that if we do this, that `InMemoryPrepopulatedDirectory` rejects the creation of character/block devices. In Buildbarn's case we don't want to allow arbitrary build actions to create character/block devices, as that can have serious security implications!

pkg/filesystem/virtual/nfsv4/nfs40_program.go
```diff
@@ -3438,6 +3472,14 @@ func nfsv4NewComponent(name string) (path.Component, nfsv4.Nfsstat4) {
 		// Inherently invalid name.
 		return path.Component{}, nfsv4.NFS4ERR_INVAL
 	}
+	if len(name) > 255 {
```
EdSchouten: I think that checks like these should live in implementations of `Directory`.

EdSchouten (commented):
Hey @visualphoenix,

First of all, glad to hear you're having fun using this code!

This code was originally designed to only serve the needs for Buildbarn. But given that it's become quite a versatile implementation (supporting macOS, Linux, and Windows) I can well imagine that people are also interested in using it outside of Buildbarn. Maybe in the long run we should move (parts of) it into a separate bb-vfs repository or something. So move the FUSE/NFSv4/WinFSP bits there, while keeping InMemoryPrepopulatedDirectory and other concrete leaf/directory implementations inside bb-remote-execution.

Anyway, let's see what we can do to get your changes in, while keeping all of Buildbarn working as intended.

pkg/filesystem/virtual/attributes.go
visualphoenix: i'll try these out - my use case is to have pjdfstest pass 100% of its tests for fs compliance, so i'll see if either of these suggestions allows for that. i'll give the IsZero() thing a shot tonight/this weekend and see how it affects my existing 100% tests passing

pkg/filesystem/virtual/attributes.go
visualphoenix: So I went back and forth on this too - a previous attempt i had made extended the size of Permissions and encoded it in there but i wasnt sure if that would be something you'd accept

visualphoenix:
Hello again @EdSchouten - I hope this message finds you well. 

Thank you again for the review on the previous diff! I appreciate the thoughtful feedback. Let me know if the cleanup here is more acceptable. I will open a new issue regarding the Permission/mode stuff. Would you like me to extend the opcode tests? Before I do I want to make sure you're okay with things.

pkg/filesystem/virtual/attributes.go
```diff
@@ -58,6 +58,12 @@ const (
 	// AttributesMaskSymlinkTarget requests the target of a symbolic
 	// link.
 	AttributesMaskSymlinkTarget
+	// AttributesMaskLastAccessTime requests the last access time
+	// (st_atim).
+	AttributesMaskLastAccessTime
```
EdSchouten: Tiny nit: could we keep the order of these attribute mask values consistent with the fields inside the Attributes struct/access methods?

pkg/filesystem/virtual/pool_backed_file_allocator.go
```diff
@@ -477,6 +477,14 @@ func (f *fileBackedFile) virtualTruncate(size uint64) Status {
...
 
 func (f *fileBackedFile) VirtualSetAttributes(ctx context.Context, in *Attributes, requested AttributesMask, out *Attributes) Status {
+	// Pool-backed regular files don't track ownership; reject chown
+	// rather than silently accepting it.
+	if _, ok := in.GetOwnerUserID(); ok {
+		return StatusErrPerm
+	}
+	if _, ok := in.GetOwnerGroupID(); ok {
+		return StatusErrPerm
+	}
```
EdSchouten, line 490: I think we also need to add this code to base_symlink_factory.go, special_file.go, and user_settable_symlink.go.

pkg/filesystem/virtual/nfsv4/nfs40_program.go
```diff
@@ -1045,49 +1063,80 @@ func (s *compoundState) opCreate(ctx context.Context, args *nfsv4.Create4args) n
 		return &nfsv4.Create4res_default{Status: st}
 	}
 
-	var attributes virtual.Attributes
+	// Decode the client-supplied create attributes once. Per the
+	// updated VirtualMknod contract the implementation owns
```
EdSchouten: Please write source code comments in such a way that it discusses the current state of affairs. So don't use phrases like the "updated [...] contract". Readers of the code should not need to care about the history behind all of this.

In fact, I don't think this comment as a whole really adds anything. Maybe better to remove entirely?

pkg/filesystem/virtual/nfsv4/nfs40_program.go
```diff
@@ -1045,49 +1063,80 @@ func (s *compoundState) opCreate(ctx context.Context, args *nfsv4.Create4args) n
...
+	// for the special-file (BLK/CHR/FIFO/SOCK) cases.
+	// VirtualMkdir / VirtualSymlink keep their current OUT-only
+	// attributes contract; createattrs for those types are silently
+	// dropped (pre-existing behaviour). Decoding upfront still gives
+	// us a fail-fast NFS4ERR_BADXDR for malformed input.
+	var createAttributes virtual.Attributes
+	if st := fattr4ToAttributes(&args.Createattrs, &createAttributes); st != nfsv4.NFS4_OK {
+		return &nfsv4.Create4res_default{Status: st}
+	}
+	var out virtual.Attributes
```
EdSchouten: Could we call this actualAttributes or something more specific?

pkg/filesystem/virtual/nfsv4/nfs40_program.go
```diff
@@ -1045,49 +1063,80 @@ func (s *compoundState) opCreate(ctx context.Context, args *nfsv4.Create4args) n
...
+		leaf, changeInfo, vs = currentDirectory.VirtualMknod(ctx, name, &createAttributes, virtual.AttributesMaskFileHandle, &out)
+		fileHandle.node = virtual.DirectoryChild{}.FromLeaf(leaf)
+	case *nfsv4.Createtype4_NF4CHR:
+		createAttributes.SetFileType(filesystem.FileTypeCharacterDevice)
+		createAttributes.SetDeviceNumber(filesystem.NewDeviceNumberFromMajorMinor(objectType.Devdata.Specdata1, objectType.Devdata.Specdata2))
+		var leaf virtual.Leaf
+		leaf, changeInfo, vs = currentDirectory.VirtualMknod(ctx, name, &createAttributes, virtual.AttributesMaskFileHandle, &out)
+		fileHandle.node = virtual.DirectoryChild{}.FromLeaf(leaf)
 	case *nfsv4.Createtype4_NF4DIR:
+		// VirtualMkdir's signature doesn't yet split create vs created
```
EdSchouten: Please don't do this. If we really want to propagate attributes to VirtualMkdir(), we should extend the signature of VirtualMkdir() to accept input and output attributes separately. The same for VirtualSymlink().

pkg/filesystem/virtual/in_memory_prepopulated_directory.go
```diff
@@ -167,6 +167,21 @@ func (c *inMemoryDirectoryContents) mayAttach(name NormalizedComponent) syscall.
 	return 0
 }
 
+// inMemoryNameMax bounds component lengths inside
```
EdSchouten: I don't see the need for making this specific change. It's perfectly fine for this directory implementation to not enforce any filename lengths, right? POSIX does not require that file systems place limits on this.

visualphoenix: Yea that’s true - I was stuck in the glibc/Linux 4096 path length mindset but you’re right - iirc it’s configurable on bsd and dynamic per filesystem there…

pkg/filesystem/virtual/nfsv4/nfs40_program.go
visualphoenix: I do really want to propagate attributes to VirtualMkdir. So basically follow the VirtualOpenChild pattern? I can do that. I'm not an expert at this, but I had a question - are you sure you really want IN/OUT for VirtualSymlink? I thought a symlink had a fixed 0777 mode. Wouldn't that mean the IN for VirtualSymlink would be empty for every impl? I kinda feel like we should keep that one as-is... Or was the idea to have symmetry generally even though the symlink syscall is just symlink(target, link)? IMHO adding IN to VirtualSymlink feels like a lie...

visualphoenix: I took a stab at this rework on the wip commit i stacked on top - i didnt bother folding any of these as fixups to the existing commits until i get feedback on the idea

EdSchouten (approved):

EdSchouten:
Hey! This looks great. Assuming CI is happy, I think we can merge the change as is. Thanks for working on this!

visualphoenix:
lmk if you want me to fix that last commit message - idk if you prefer to squash or merge commit - i can clean that `wip` commit up to have a better name if you so desire

EdSchouten:
Don't worry about it. My plan was to squash it, so I'll make some editorial mixups if needed.

visualphoenix:
aww crud - i forgot the test files when i added the securityFlavors param

EdSchouten:
Whoops! Looks like the build is still failing.

visualphoenix:
will fix - 1s

visualphoenix:
Sorry - just got bazel working - got a few more tests to fix.

visualphoenix:
its going to fail that run

visualphoenix:
i have to add EXPECT stuff for returning StatusErrPerm for VirtualSetAttributes for FUSE and VirtualMknod for NFS - and i have to fix the encoder tests because i modified the mask adding LastAccessTime and LastStatusChangeTime - i'll have to do these tomorrow night. sorry for overlooking these. i'll make sure i dont bug you until i get the tests working locally - i'll remove the WIP from the PR when its done. apologies again.

visualphoenix:
Ready for review
```
INFO: 1443 processes: 2157 action cache hit, 274 internal, 1169 processwrapper-sandbox.
...
Executed 15 out of 19 tests: 19 tests pass.
```

pkg/filesystem/virtual/stateless_handle_allocating_cas_file_factory_test.go
```diff
@@ -124,16 +125,16 @@ func TestStatelessHandleAllocatingCASFileFactory(t *testing.T) {
...
+			DoAndReturn(func(ctx context.Context, buf []byte, off uint64) (int, bool, virtual.Status) {
 				copy(buf, "Hello")
 				return 5, false, virtual.StatusOK
 			}).
 			Times(10)
 
 		for i := 0; i < 10; i++ {
 			var buf [5]byte
-			n, eof, s := monitoringLeaf.VirtualRead(buf[:], 0)
+			n, eof, s := monitoringLeaf.VirtualRead(context.Background(), buf[:], 0)
```
EdSchouten: Nit: This could also use `ctx` if the gomock initialization at the top used WithContext().

EdSchouten (approved):
One small remark. Be sure to fix, and I'll merge assuming CI is happy.

visualphoenix:
apologies for not fixing the lint fail at the same time. at this point i expect all of the ci checks to pass. thanks again for the detailed and excellent code review! <3

PR #237, opened 2026-06-23
Add copy fallback when hardlinking hits the filesystem link limit

### LINKED ISSUE #236: native build directory: copy fallback when hardlink hits the filesystem link limit (NTFS 1023) on Windows
### Problem

On Windows workers using the `native` build directory, `bb_worker` hardlinks cached CAS blobs from the local cache into each action's input root (`hardlinkingFileFetcher`). NTFS caps a single file at **1023 hard links**. A *hot* shared blob — e.g. a common toolchain input needed by many actions — gets hardlinked into more than 1023 input roots and `CreateHardLink` fails with `ERROR_TOO_MANY_LINKS` (0x476). buildbarn wraps this as `codes.Internal`, so the action aborts mid-build:

```
Failed to create hardlink to cached file "...":
An attempt was made to create more links on a file than the file system supports.
```

This is reproducible with a fan-out build (many actions sharing one large toolchain input) on a worker whose `C:\worker` cache/build directories are on NTFS.

### Why `native` (and not the virtual filesystem)

The `virtual` build directory avoids hardlinks, but on Windows its backend is WinFSP, which requires installing the WinFSP **kernel-mode driver** on the worker host. On managed Kubernetes (e.g. AKS Windows node pools) you cannot install a kernel driver on the node, and the worker base image (`nanoserver`) has no driver. So `native` is the only viable build directory for these deployments, and the link cap is a hard wall there.

### Proposal

When the hardlink fails because the file has reached the filesystem's maximum link count (`ERROR_TOO_MANY_LINKS` on Windows, `EMLINK` on POSIX), fall back to **copying** the cached file into the input root instead of failing. A copy is an independent inode with its own link count, so materialization succeeds; only the few over-limit hot blobs pay the copy cost, and correctness is unchanged because the input root only needs the file's contents.

### Status

I have a working patch (small, single-module — no `bb-storage` change needed), with a unit test, that passes `bazel test //pkg/cas:cas_test`, the `windows_amd64` cross-build, golint, and the reformat/gofmt gate locally. Happy to open a PR if you're open to this approach.

Would you accept this, or do you see a preferred way to handle the link cap on Windows `native` workers?

ysmaoui:
Fixes #236.

### Problem

When `bb_worker` uses a `native` build directory, `hardlinkingFileFetcher` hardlinks cached CAS blobs from the local cache into each action's input root. Filesystems cap the number of hard links per file — NTFS allows **1023**. A file shared by a large number of input roots (e.g. a common toolchain input on a high-fan-out build) exceeds the cap, and the hardlink fails with `ERROR_TOO_MANY_LINKS` on Windows (`EMLINK` on POSIX). This is wrapped as `codes.Internal` and the action aborts:

```
Failed to create hardlink to cached file "...":
An attempt was made to create more links on a file than the file system supports.
```

On Windows the `virtual` (WinFSP) build directory avoids hardlinks, but it needs the WinFSP kernel driver installed on the host — not possible on managed Kubernetes (e.g. AKS Windows node pools) — so `native` is the only option there and the link cap is a hard wall.

### Change

When the hardlink fails because the cached file has reached the filesystem's maximum link count, fall back to **copying** the file into the input root instead of failing. The copy is an independent inode with its own link count, so materialization succeeds. Only the few over-limit hot blobs pay the copy cost; correctness is unchanged because the input root only needs the file's contents, not a shared inode.

The fix is contained to `pkg/cas` (no `bb-storage` change). `isHardlinkLimitReached` is platform-split: `ERROR_TOO_MANY_LINKS` on Windows, `EMLINK` elsewhere.

### Testing

- New unit test `TestHardlinkingFileFetcherCopyFallbackOnLinkLimit` exercises the fallback (via `EMLINK`); `bazel test //pkg/cas:cas_test` passes.
- `windows_amd64` cross-build of `//cmd/bb_worker` passes.
- golint and the reformat/gofmt gate are clean.

### DIFF
--- a/pkg/cas/BUILD.bazel
+++ b/pkg/cas/BUILD.bazel
@@ -7,11 +7,14 @@ go_library(
         "blob_access_file_fetcher.go",
         "caching_directory_fetcher.go",
         "configuration.go",
+        "copy_on_link_limit.go",
         "decomposed_directory_walker.go",
         "directory_fetcher.go",
         "directory_walker.go",
         "file_fetcher.go",
         "hardlinking_file_fetcher.go",
+        "link_limit_other.go",
+        "link_limit_windows.go",
         "suspending_directory_fetcher.go",
     ],
     importpath = "github.com/buildbarn/bb-remote-execution/pkg/cas",
@@ -32,7 +35,12 @@ go_library(
         "@org_golang_google_grpc//status",
         "@org_golang_google_protobuf//encoding/protowire",
         "@org_golang_google_protobuf//proto",
-    ],
+    ] + select({
+        "@rules_go//go/platform:windows": [
+            "@org_golang_x_sys//windows",
+        ],
+        "//conditions:default": [],
+    }),
 )
 
 go_test(
@@ -51,6 +59,7 @@ go_test(
         "@com_github_buildbarn_bb_storage//pkg/blobstore/slicing",
         "@com_github_buildbarn_bb_storage//pkg/digest",
         "@com_github_buildbarn_bb_storage//pkg/eviction",
+        "@com_github_buildbarn_bb_storage//pkg/filesystem",
         "@com_github_buildbarn_bb_storage//pkg/filesystem/path",
         "@com_github_buildbarn_bb_storage//pkg/testutil",
         "@com_github_stretchr_testify//require",
--- a/pkg/cas/copy_on_link_limit.go
+++ b/pkg/cas/copy_on_link_limit.go
@@ -1,0 +1,44 @@
+package cas
+
+import (
+	"io"
+
+	"github.com/buildbarn/bb-storage/pkg/filesystem"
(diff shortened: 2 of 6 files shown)


EdSchouten (changes requested):
Though it obviously makes sense to fix this issue, I'm not happy with the overall strategy that is chosen. Because now if we hit the limit, then for this particular file we will **always** create copies. So if there is a need to instantiate a file 2000 times, we'll have 1 file with link count 1023, and 977 files with link count 1.

My recommendation would be to do the following if we run into `EMLINK` issues:

1. Copy the file from the hard linking cache to the target location.
2. Atomically rename the file from the target location into the hard linking cache. This causes the existing file in the cache to be replaced with an instance that has a link count of 1.
3. Attempt to create a hard link from the cache to the target location again.

PR #238, opened 2026-06-26
Support FUSE on FreeBSD

moroten:
This PR contains two distinct commits:
* Upgrade go-fuse to include FreeBSD support.
* Add FreeBSD support for the virtual FUSE file system.

This has been tested in a deployment and works.


b3cramer (approved):
Thanks so much for pulling this together. Working great.

EdSchouten:
Looking good!

Can you update comments in https://github.com/buildbarn/bb-remote-execution/blob/main/pkg/proto/configuration/filesystem/virtual/virtual.proto#L20 to mention FreeBSD is also supported? Thanks!

EdSchouten:
Hey @moroten, @b3cramer, PTAL @ https://github.com/buildbarn/bb-remote-execution/pull/239

moroten:
> Can you update comments in https://github.com/buildbarn/bb-remote-execution/blob/main/pkg/proto/configuration/filesystem/virtual/virtual.proto#L20 to mention FreeBSD is also supported? Thanks!

I was too slow, lets use #239 instead.

PR #243, opened 2026-07-30
Add file system access profile to execute response

osjo-meroton:
Currently bb-portal uses the latest file system access profile saved to the FSAC. This profile may differ from the profile produced by the action being viewed, for example when viewing old action results. With this change, action result contains its corresponding access profile.

### DIFF
--- a/pkg/builder/prefetching_build_executor.go
+++ b/pkg/builder/prefetching_build_executor.go
@@ -152,11 +152,13 @@ func (be *prefetchingBuildExecutor) Execute(ctx context.Context, filePool pool.F
 		return nil
 	})
 
-	if err := group.Wait(); err == nil {
+	err = group.Wait()
+	newProfile := be.computeProfile(bloomFilterMonitor)
+	if err == nil {
 		// Action completed successfully. Store an updated
 		// profile in the File System Access Cache if it is
 		// different from the one fetched previously.
-		if newProfile := be.computeProfile(bloomFilterMonitor); !proto.Equal(existingProfile, newProfile) {
+		if !proto.Equal(existingProfile, newProfile) {
 			if err := be.fileSystemAccessCache.Put(ctx, reducedActionDigest, buffer.NewProtoBufferFromProto(newProfile, buffer.UserProvided)); err != nil {
 				attachErrorToExecuteResponse(response, util.StatusWrap(err, "Failed to store file system access profile"))
 			}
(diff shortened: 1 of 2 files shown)


EdSchouten (commented):
The problem is that the profile can become pretty big. I don't think it's a good idea to attach tens of kilobytes of data.

moroten:
> The problem is that the profile can become pretty big. I don't think it's a good idea to attach tens of kilobytes of data.

The recommended maximum size is 64 kB, so yes it is quite a bit of data. The goal is to get the access profile for a specific action. One could store the profile in FSAC for both the reduced action digest and the full action digest. That would though fill up the FSAC and there is race condition between execution and access if multiple actions are run in parallel.

What about storing the profile in the CAS and reference it from the action result? Then it's just ~70 bytes of data.

EdSchouten:
> What about storing the profile in the CAS and reference it from the action result? Then it's just ~70 bytes of data.

Yeah, that makes more sense, though I would recommend gating it by a configuration option. What are your thoughts on including it in `ExecuteResponse.server_logs`? Or do you really need it to be part of `ActionResult`?

osjo-meroton:
Gating it by a configuration option seems very sensible.

As for using the logs instead. We would need the profile to be easily accessible at runtime.

moroten:
> As for using the logs instead. We would need the profile to be easily accessible at runtime.

We meant [`remoteexecution.ExecutionResponse.server_logs`](https://github.com/bazelbuild/remote-apis/blob/becdd8f9ff811df88a22d3eadd6341753d51d167/build/bazel/remote/execution/v2/remote_execution.proto#L1624), not to be confused with any logs of stdout/stderr on the servers.

osjo-meroton:
Has been rebased onto main and resubmitted with suggested changes. It now uses the Server Logs and CAS to log the file access profiles.

pkg/proto/configuration/bb_worker/bb_worker.proto
```diff
@@ -475,6 +475,16 @@ message PrefetchingConfiguration {
   // Was 'download_concurrency'. This option has been moved to
   // 'input_download_concurrency' in ApplicationConfiguration.
   reserved 4;
+
+  // Enables logging of generated file system access profiles.
+  //
+  // The profile digest is attached to the response Server Logs under
+  // "file-access-profile". The profile itself is also stored to
```
EdSchouten: Can we call this file `buildbarn.fsac.FileSystemAccessProfile`, just like the fully-qualified name of the Protobuf message?

osjo-meroton: Done

EdSchouten (approved):
LGTM! Will merge if CI is happy.

PR #244, opened 2026-08-05
Stamp image tags from STABLE_ status keys

matx-sjacob:
Related https://github.com/buildbarn/bb-storage/issues/358

`BUILD_SCM_REVISION` and `BUILD_SCM_TIMESTAMP` have no `STABLE_` prefix, so Bazel files them in `volatile-status.txt` and reports an unchanging digest for that file whatever it contains. The `expand_template` in bb-storage that turns them into the image tag is therefore never invalidated, and `_stamped.tags.txt` is generated once and reused indefinitely — on a build machine whose cache persists, every later build pushes its image to the tag produced by the *first* build on that machine.

The bb-storage bump is part of the fix, not incidental. That template is what expands these keys, and the pin here (`3991d6e`) predates `33530b6`, which switched it to the prefixed names. Renaming the keys alone would leave the pinned template still reading the volatile ones, so nothing would change.


EdSchouten:
I think you also ought to bump the dependency on bb-storage to make this work.

matx-sjacob:
> I think you also ought to bump the dependency on bb-storage to make this work.

Cool, that's a better solution, I've made it part of this PR. PTAL when you can @EdSchouten

EdSchouten:
Please also bump the hash of bb-storage in go.mod. I know the Bazel build won't use it, but it's necessary to make plain "go build" work.

matx-sjacob:
Done. PTAL @EdSchouten  - Thanks!

EdSchouten:
Interesting. Why doesn't GitHub try to run CI jobs for this PR? Will merge this change once I've confirmed this doesn't cause any regressions.

PR #246, opened 2026-08-12
Clean up Windows action process trees

### LINKED ISSUE #245: Windows: remote executor fails to delete build root when actions leave subprocesses running
On Windows, an action can (and very often will) spawn a subprocess and exit without terminating it. The remote executor then attempts to delete the action root, but the surviving subprocess may still hold files or its working directory open inside that root.

This causes cleanup to fail with an error stating that the root directory cannot be deleted because it is still in use.

### Reproduction

1. Run an action that starts a long-lived child process.
2. Have the parent process exit without waiting for or terminating the child.
3. Ensure the child keeps its working directory inside the action root or holds a file there open.
4. Observe that action cleanup fails when the executor tries to delete the root.

The issue is reproducible reliably on Windows and will accur pretty much always for some of our tools.

### Suggested fix

[Bazel’s native Windows launcher](https://github.com/bazelbuild/bazel/blob/master/src/main/native/windows/process.cc) handles this with a Windows Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`. It starts the root process suspended, assigns it to the non-breakaway job, and only then resumes it. This prevents the process from spawning descendants before job assignment.

The approach cannot be translated one-to-one when using Go’s `os/exec`. Bazel receives and retains the primary thread handle from `CreateProcessW`, allowing it to call `ResumeThread` directly. [Go’s Windows process implementation](https://go.dev/src/syscall/exec_windows.go) closes that thread handle before `cmd.Start()` returns and exposes only the process and PID. A Go implementation must therefore use another way to resume the suspended process, such as enumerating its threads, or replace `os/exec` process creation with lower-level Windows API calls.

When the action finishes or is cancelled, closing or terminating the job kills all surviving descendants. A completion port can be used to wait for `JOB_OBJECT_MSG_ACTIVE_PROCESS_ZERO` before deleting the action root, ensuring no process still holds files or directories open.

### (Maybe to discuss)

Honestly, I would just move the windows implementation always to this object group as it is common in bazel. However, as there is the `clean_process_table` for linux and mac one could argue that we want that for Windows as well. But then it should be more general like a `process_tree_cleanup` or `cleanup_action_processes`. I think this discussion will quickly become a scope creep so I would just roll it out for windows out of the box.

alex-the-third:
On Windows, actions may exit while leaving subprocesses running. Those subprocesses can retain their working directory or open files inside the action root, causing the executor to fail when removing it after the action completes.

This implements the approach proposed in #245 and follows [Bazel's native Windows launcher](https://github.com/bazelbuild/bazel/blob/master/src/main/native/windows/process.cc).

## Implementation

Each action process is placed in its own non-breakaway Windows Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`.

The root process is created suspended, assigned to the job, and only then resumed. This prevents it from spawning descendants before job assignment. Because Go's `os/exec` closes the primary thread handle before `cmd.Start()` returns, the implementation locates the root process's threads by PID and resumes them through the Windows API.

When the root process exits, any remaining processes in the job are terminated. The runner then waits for `JOB_OBJECT_MSG_ACTIVE_PROCESS_ZERO` before returning, ensuring that descendants no longer hold files or directories inside the action root.

Command cancellation terminates the entire job instead of only the root process. Startup cancellation is serialized with process resumption so cancellation cannot terminate the suspended process between the cancellation check and resume.

Non-Windows behavior remains unchanged.

## Testing

Added Windows tests covering:

- A parent that exits while leaving a child holding a file open inside the action root.
- Cancellation of a running parent and descendant.
- The race between the startup cancellation check and resuming the suspended process.
- Immediate removal of the action root after `Run()` returns.

Fixes #245

### DIFF
--- a/pkg/runner/local_runner_test.go
+++ b/pkg/runner/local_runner_test.go
@@ -3,11 +3,13 @@ package runner_test
 import (
 	"context"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
 	"testing"
+	"time"
 
 	"github.com/buildbarn/bb-remote-execution/internal/mock"
 	"github.com/buildbarn/bb-remote-execution/pkg/proto/resourceusage"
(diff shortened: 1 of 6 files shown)


EdSchouten:
Though I appreciate the work that went into this change, I think that this is not something we should try to tackle on the Buildbarn side. In my opinion it is a feature that should be supported natively by Go's `os/exec` API. Just like how it supports `chroot()` on UNIX. In fact, there is currently work under way to get this added:

https://github.com/golang/go/issues/79927
https://github.com/golang/go/issues/80415

Can you please work with the authors of those issues to validate that all of the features you need are part of those changes? Thanks!

pkg/runner/local_runner_test.go
```diff
@@ -549,3 +551,203 @@ func TestLocalRunnerRun(t *testing.T) {
...
+		if value, ok := os.LookupEnv(name); ok {
+			environmentVariables[name] = value
+		}
+	}
+
+	runner := runner.NewLocalRunner(buildDirectory, buildDirectoryPathBuilder, runner.NewPlainCommandCreator(&syscall.SysProcAttr{}), false)
+	response, err := runner.Run(context.Background(), &runner_pb.RunRequest{
+		Arguments: []string{
+			testBinaryPath,
+			"-test.run=TestLocalRunnerRunWindowsSubprocessCleanupHelper",
```
EdSchouten: I don't think it's safe to assume that you can do this. What happens if rules_go's go_test() is restructured to compile Go test binaries with a different type of entry point?

alex-the-third: Hmm, agree that we should probably keep the implementation details of rules_go out of here, though I doubt that it is likely to change...
Anyhow, I could either move the helper dispatch into `TestMain` as in (rules_go #4335/#4336) or use a dedicated `go_binary`.

alex-the-third:
I tested Buildbarn against the actual implementation in Go [CL 801640](https://go-review.googlesource.com/c/go/+/801640). It satisfies the launch-time containment requirements and lets me remove `CREATE_SUSPENDED`, post-start `AssignProcessToJobObject`, thread enumeration/resumption, and the associated startup cancellation synchronization.
The Windows implementation still works as desired but as you suspected can be done a lot leaner with it.

The remaining code is the Buildbarn-specific action lifecycle: creating/configuring the per-action Job Object, whole-tree cancellation, terminating descendants after the root exits, waiting for `ACTIVE_PROCESS_ZERO`, and cleanup.
I think this is acceptable because to me those are executor semantics and not generics.

I found one Go API edge case (`Jobs: []Handle{}` gives `ERROR_BAD_LENGTH`, unlike nil) and still need clarification.

You can close/defer this PR until Go is ready - we are not pressured by time here and I just continue using my patched executor. 

Also if you want I can publish a prototype of the stuff that would remain after the Go API change.

EdSchouten:
Awesome! Yeah, if you want, you can just push the eventual changes in this PR and switch it back to draft. Then by the time Go 1.28 is out, all you need to do is move it out of draft again.

Thanks for looking into this!

alex-the-third:
Done. Should someone need the workaround until then, they can build  https://github.com/buildbarn/bb-remote-execution/pull/246/commits/65f1afdf6075ac8d61e1da668f71bafce4caf04b.

PR #248, opened 2026-08-18
Update to bb-storage without a CAS blobAccess

meroton-benjamin:
This commit updates to bb-storage that doesn't export the Content Addressable Storage (CAS) as a blobstore.BlobAccess. The CAS is instead built from two low level storage primitives. The Chunk Storage (CS) which contains raw chunks of data and the Chunk List Storage (CLS) which contains lists that describes which chunks make up a blob in the CAS.

This had a fairly large impact on bb-remote-execution as many components assumed that it would interact with the CAS via a blobstore.BlobAccess but this has now been replaced with a cdc.ContentAddressableStorage.


pkg/cas/existence_precondition_content_addressable_storage.go
```diff
@@ -0,0 +1,68 @@
...
+// code "FAILED_PRECONDITION" instead of "NOT_FOUND" for Get() style
+// operations. This is used by worker processes to make
+// Execution::Execute() comply to the protocol.
+func NewExistencePreconditionContentAddressableStorage(contentAddressableStorage cdc.ContentAddressableStorage) cdc.ContentAddressableStorage {
+	return &existencePreconditionContentAddressableStorage{
+		ContentAddressableStorage: contentAddressableStorage,
+	}
+}
+
+func (cas *existencePreconditionContentAddressableStorage) FetchChunk(ctx context.Context, d digest.Digest) ([]byte, error) {
```
EdSchouten, line 31: Hmmm... This brings up an interesting question. In case of chunking, what digest should be used as part of `FAILED_PRECONDITION` errors? I would argue that you always want to use the digest of the containing object, not the chunk. Clients may be oblivious of chunking. And there's also no guaranteed/fully normalized form of chunking, so returning the digest of the full file is the only meaningful thing you can do here. If that's the case, then there is no way we can convert the error at this specific level/layer.

This is probably worth raising in the working group as well?

pkg/builder/caching_build_executor.go
```diff
@@ -12,6 +12,7 @@ import (
 	re_util "github.com/buildbarn/bb-remote-execution/pkg/util"
 	"github.com/buildbarn/bb-storage/pkg/blobstore"
 	"github.com/buildbarn/bb-storage/pkg/blobstore/buffer"
+	"github.com/buildbarn/bb-storage/pkg/blobstore/cdc"
```
EdSchouten: I think that this is a bit of an odd package for having the ContentAddressableStorage interface. Why can't we just call it pkg/cas?

pkg/builder/caching_build_executor_test.go
```diff
@@ -159,15 +160,18 @@ func TestCachingBuildExecutorCachedSuccessNonZeroExitCode(t *testing.T) {
...
+	contentAddressableStorage.EXPECT().PutChunk(
 		ctx,
 		digest.MustNewDigest("freebsd12", remoteexecution.DigestFunction_SHA256, "bb1107706f3aa379d68aa61062f56d99d24a667ec18d5756fb6df1ba9baa1fdc", 93),
 		gomock.Any(),
 	).
-		DoAndReturn(func(ctx context.Context, digest digest.Digest, b buffer.Buffer) error {
-			historicalExecuteResponse, err := b.ToProto(&cas_proto.HistoricalExecuteResponse{}, 10000)
-			require.NoError(t, err)
+		DoAndReturn(func(ctx context.Context, digest digest.Digest, b []byte) error {
+			historicalExecuteResponse := testutil.MustUnmarshal(t, b, &cas_proto.HistoricalExecuteResponse{})
```
EdSchouten: My suggestion would be to simply write this as:
```go
var historicalExecuteResponse cas_proto.HistoricalExecuteResponse
require.NoError(t, proto.Unmarshal(&historicalExecuteResponse))
```
It's not significantly longer, but has the advantage that we have strong typing (and don't need to resort to generics).

pkg/builder/local_build_executor.go
```diff
@@ -231,12 +234,15 @@ func (be *localBuildExecutor) Execute(ctx context.Context, filePool pool.FilePoo
 		attachErrorToExecuteResponse(response, util.StatusWrap(err, "Failed to extract digest for command"))
 		return response
 	}
-	commandMessage, err := be.contentAddressableStorage.Get(ctx, commandDigest).ToProto(&remoteexecution.Command{}, be.maximumMessageSizeBytes)
+	if commandDigest.GetSizeBytes() > int64(be.maximumMessageSizeBytes) {
+		attachErrorToExecuteResponse(response, status.Errorf(codes.InvalidArgument, "Command is %d bytes which exceeds the maximum message size byte of %d bytes", commandDigest.GetSizeBytes(), be.maximumMessageSizeBytes))
+		return response
+	}
+	command, err := cdc.GetProto(ctx, be.contentAddressableStorage, commandDigest, &remoteexecution.Command{})
```
EdSchouten: What are your thoughts on simply letting `GetProto()` take the `maximumMessageSizeBytes`?

pkg/builder/naive_build_directory.go
EdSchouten: In the pre-CDC era, this overall design for UploadFile() sort of made sense. First fully hash the file, and then create a `buffer.Buffer` out of it. However, with CDC this feels a bit odd. I don't know how to express it just yet; let me think about this a bit more. I guess I'll also need to read up on the bb-storage changes first.

pkg/builder/noop_build_executor_test.go
```diff
@@ -174,8 +192,8 @@ func TestNoopBuildExecutor(t *testing.T) {
 	t.Run("SuccessCustomTemplate", func(t *testing.T) {
 		// If a custom template is provided in the environment,
 		// it should be preferred over the default template.
-		contentAddressableStorage.EXPECT().Get(ctx, digest.MustNewDigest("build", remoteexecution.DigestFunction_SHA256, "9da17cb226048f5bb3e6a20311b551e73ce8ac0d408e69e737d28a8f3179d0ce", 300)).
-			Return(buffer.NewProtoBufferFromProto(&remoteexecution.Command{
+		protoBytes, err := proto.Marshal(
```
EdSchouten: It's a bit of a shame we need to write proto.Marshal()/proto.Unmarshal() more often in our tests now. Not that I know how to fix this exactly.

EdSchouten (commented):
Will continue reviewing later!

pkg/builder/naive_build_directory.go
```diff
@@ -188,25 +188,24 @@ func (d *naiveBuildDirectory) UploadFile(ctx context.Context, name path.Componen
 	// used to compute the digest. This ensures uploads succeed,
 	// even if more data gets appended in the meantime. This is not
 	// uncommon, especially for stdout and stderr logs.
-	if err := d.options.contentAddressableStorage.Put(
+	if err := d.options.blobUploader.UploadBlob(
```
EdSchouten:
```go
UploadBlob(ctx context.Context, r filesystem.FileReader) (digest.Digest, error)
```
?

pkg/cas/blob_uploader.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+
+	"github.com/buildbarn/bb-storage/pkg/digest"
+)
+
+// BlobUploader is an interface for uploading an arbitrary blob.
+type BlobUploader interface {
+	UploadBlob(ctx context.Context, d digest.Digest, blob Blob) error
```
EdSchouten: - Let this take a filesystem.FileReader
- Instead of digest.Digest, use digest.Function.
- Let this also return digest.Digest...?

meroton-benjamin:
Replaced with #252 as github does not allow pull requests to modify their source/destination branches.

PR #249, opened 2026-08-22
Consume the storage.MessageReader interface

meroton-benjamin:
When fetching messages from the CAS, do it via the storage.MessageReader interface. This completely decouples the scheduler.InMemoryBuildQueue and builder.noopBuildExecutor from the blobstore.BlobAccess interface.


meroton-benjamin:
Closed in favor of #251

PR #252, opened 2026-09-06
Update to bb-storage without a CAS blobAccess

meroton-benjamin:
This commit updates to bb-storage that doesn't export the Content Addressable Storage (CAS) as a blobstore.BlobAccess. The CAS is instead built from two low level storage primitives. The Chunk Storage (CS) which contains raw chunks of data and the Chunk List Storage (CLS) which contains lists that describes which chunks make up a blob in the CAS.

This had a fairly large impact on bb-remote-execution as many components assumed that it would interact with the CAS via a blobstore.BlobAccess but this has now been replaced with a cdc.ContentAddressableStorage.


meroton-benjamin:
Depends-On: https://github.com/buildbarn/bb-storage/pull/366

PR #256, opened 2026-09-16
fix(runner): Don't use procfs ctime for process cleaning.

mkosiba:
The process table cleaner kills all processes started after bb_runner started. It does so by looking at the `/proc/<pid>` ctime. Memory pressure can cause these files to be GC'd and created anew with newer timestamps resulting in the runner killing itself. Unfortunately parsing process start time from `/proc/<pid>/stats` is [complicated](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L835), so instead we rely on a snapshot of currently running process IDs taken when the runner starts. This is a bit less reliable as PIDs can be reused, but should be sufficient for most users.


cmd/bb_runner/main.go
```diff
@@ -91,19 +90,34 @@ func main() {
 
 		// Kill processes that actions leave behind by daemonizing.
 		// Ensure that we only match processes belonging to the current
-		// user that were created after bb_runner is spawned, as we
+		// user that did not yet exist when bb_runner was spawned, as we
 		// don't want to kill unrelated processes.
+		//
+		// This does assume that processes that were already running at
+		// startup outlive bb_runner.
```
EdSchouten, line 97: I don't think that this is a desirable assumption.

PR #257, opened 2026-09-18
Bypass hardlink caching for empty files

dzbarsky:
Fetch empty files directly instead of sharing one cached inode. Input trees with many empty files can otherwise exhaust the filesystem's hardlink limit; empty blobs do not need a download. This also avoids cache bookkeeping for those files.

Added coverage for repeated executable and non-executable empty files. The focused `TestHardlinkingFileFetcher` test fails before the change and passes afterward in an isolated copy with generated mocks and a Darwin `O_SEARCH` dependency shim. The unmodified checkout's Go test command is blocked by those existing dependency/generation requirements.


EdSchouten:
Any type of file can currently exhaust the filesystem's hard link limit. Not just empty files. A change like this is therefore a workaround, which doesn't actually solve the problem at hand.

Somebody sent out a PR to address this issue some time ago, but didn't respond to feedback I had left: https://github.com/buildbarn/bb-remote-execution/pull/237.

PR #258, opened 2026-09-20
Pass cmd.exe invocations a raw command line on Windows

ysmaoui:
On Windows, any action whose command line contains a double quote reaches the
worker corrupted, so a quoted path resolves against the root of the current
drive instead of the working directory. Quoting a path in a genrule is normal,
and Bazel emits quoted paths in the `cmd.exe` command lines it generates, so
every Windows build that has one is affected.

It is hard to attribute: the action fails with `The system cannot find the path
specified` naming a path that exists, which reads as a missing input rather than
as a worker problem.

### Mechanism

`NewPlainCommandCreator` sets `cmd.Args` but never `SysProcAttr.CmdLine`, and on
Windows `SysProcAttr` is empty, so `syscall.StartProcess` falls back to
`makeCmdLine(argv)`. That applies `CommandLineToArgvW` escaping, turning a
literal `"` into `\"`. `cmd.exe` does not parse by those rules.

Bazel passes a genrule script as one argument:

    cmd.exe /S /E:ON /V:ON /D /c  copy "out\libfoo.a" "out\FOO.a" >NUL

Go escapes that to:

    cmd.exe /S /E:ON /V:ON /D /c "copy \"out\libfoo.a\" \"out\FOO.a\" >NUL"

`/S` makes `cmd.exe` drop the first and last quote and pass the rest through
untouched. Backslash is not an escape character to it, so `copy` receives
`\out\libfoo.a`: a leading backslash, hence the error naming a path rather than
a file.

### Fix

Set `SysProcAttr.CmdLine` explicitly for the one argument shape where Go's
escaping is wrong, and leave every other shape alone. Go's escaping is correct
for programs that split their command line with `CommandLineToArgvW`, which is
nearly all of them; `cmd.exe` is the documented exception.

The rewrite applies only when argv[0]'s base name is `cmd.exe`, an argument
equals `/c`, exactly one argument follows it, `/S` is present, and nothing
before `/c` would need escaping. The single-argument condition matters: quoting
is what separates multiple arguments after `/c`, so merging them would change
how the invoked program sees them. `/S` is checked rather than added, because
the unconditional quote stripping this relies on is the caller's choice.

Bazel hit the same defect in its local Windows execution and fixed it the same
way, in 47e832a (#12530, for #11975), by not escaping arguments when argv[0] is
`cmd.exe`. Bazel's rule is broader than this one: it stops escaping entirely for
that program, giving up spaces in arguments to programs launched through
`cmd.exe`. That is a reasonable trade when you also generate the argv, but a
worker serves arbitrary clients, so this change declines the shapes it cannot
rewrite safely. Happy to widen it to match Bazel if you prefer that.

Note that `NewPlainCommandCreator` hands the same `*syscall.SysProcAttr` to
every action, so the struct is copied before `CmdLine` is set; writing it in
place would publish one action's command line to the others.

REAPI does not say how `Command.arguments` maps onto a Windows command line, so
every Windows worker has to pick a rule. If you would rather settle that
differently, I am glad to follow.

### Testing

Unit tests cover the Bazel argument shape and the shapes that must be left
alone. Built from this change and run on a Windows Server 2022 worker: a genrule
with `cmd_bat = 'copy "$<" "$@" >NUL'` fails on an unpatched runner and builds
on a patched one, and a `dir` of the same path, quoted, succeeds in the same
action.

`go test ./pkg/runner/...` cannot run outside Bazel in a clean checkout, because
the tests import `//internal/mock`, whose sources are Gazelle-generated; the new
test was also run in a standalone module to confirm it has no build-tag or
platform dependency. The `go_test` rule moves from `deps` to `embed` because the
function under test is unexported and every existing test in the package is
external.


EdSchouten:
> REAPI does not say how `Command.arguments` maps onto a Windows command line, so
> every Windows worker has to pick a rule. If you would rather settle that
> differently, I am glad to follow.

Please work with the remote API working group to get the REv2 spec updated to accurately describe what the intended semantics on Windows are supposed to be. I don't think we should make any changes along these lines on the Buildbarn side without having that sorted out.

ysmaoui:
That's fair, and I'm not arguing the spec gap away.

But I don't think `bb_runner` is currently abstaining on an undefined question. It already picked a rule, by way of Go's default (`SysProcAttr.CmdLine` empty, so `StartProcess` escapes argv for `CommandLineToArgvW`), and that rule disagrees with Bazel. Bazel hit the same problem in its own local execution five years ago ([#11975](https://github.com/bazelbuild/bazel/issues/11975), [#12530](https://github.com/bazelbuild/bazel/pull/12530)) and still does it that way in [`WindowsSubprocessFactory`](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/shell/WindowsSubprocessFactory.java#L69): when `argv[0]` is `cmd.exe`, skip the escaping.

So a genrule with a quoted path builds locally and fails on Buildbarn. The error is `The system cannot find the path specified`, naming a path that is plainly there, which reads as a missing input and sends you looking somewhere else entirely.

This PR moves `bb_runner` onto Bazel's rule rather than inventing a third one. If you'd still rather not carry something that isn't written down, would a config option work, default off? Otherwise I'm happy to close this and revisit once REv2 covers it.

EdSchouten:
Please work with the remote API working group to get the spec updated first. Feel free to leave this PR open in the meantime.

PR #259, opened 2026-09-22
Replace all references to bb-browser with bb-portal

EdSchouten:
We just pushed a change to fully deprecate and archive bb-browser by the end of this year. The browser_url configuration option accepted by the services in this repository can also just port to bb-portal. It therefore makes more sense to rename these options to portal_url.

Cc @oscar-meroton


pkg/proto/configuration/bb_scheduler/bb_scheduler.proto
meroton-benjamin, line 131: Do we want to add a deprecation notice on the admin_http_servers as well? BB-portal is already a ui which interacts with the scheduler via the above mentioned bqs api.

meroton-benjamin (approved):
LGTM but do we want to mention bb-portal and/or deprecate the admin http api for the scheduler?

pkg/proto/configuration/bb_scheduler/bb_scheduler.proto
EdSchouten: Yep! I'll send out a PR for that right after merging this one. This is just about deprecating bb-browser; not deprecating the scheduler web UI.

PR #261, opened 2026-09-22
Deprecate bb_scheduler's web UI, so we can remove it on 2027-07-01

EdSchouten:
bb-portal is capable of displaying exactly the same information nowadays, so there is no need to have an integrated web UI.


meroton-benjamin (approved):
👍

